PR #156, opened 2014-07-29
WIP: Modifications to sugar toolkit for working with social-help activity

prasoon2211:
This PR contains changes that were required to make the social help activity work with sugar. There are two things we're doing:
1. Checking whether the currently running activity has a discussion available.
2. Adding a button to the activity toolbar for launching social-help based on the result of 1.

Both of the calls are implemented via a dbus callable method which I've [added in another PR](https://github.com/sugarlabs/sugar/pull/391/).

PS: I still have to run pep8 and pyflakes on the changes.


samdroid-apps:
Merged a different patch

PR #180, opened 2014-12-28
Removing unwanted vertical padding in the alignment

i5o:
Fixes SL#4792

GCI task. https://www.google-melange.com/gci/task/view/google/gci2014/5857652508721152


samdroid-apps:
Removing the padding doesn't fix this bug.  It actually looks exactly the same as the screenshot in the bug with and without this patch applied.

Also @godiard I had a look into this.  Please see the comments on the ticket.  Let's continue the discussion there.

PR #183, opened 2015-01-02

SAMdroid-apps:
This patch uses idle_add so it runs in the background.  When it is forced to run in the foreground it only takes about 0.6s (on Core2 @2GHz) to get all the mimes.

GCI Task:  https://www.google-melange.com/gci/task/view/google/gci2014/5810064153640960


SAMdroid-apps:
Replaced!

PR #184, opened 2015-01-03
Log error when the espeak plugin is not installed.

ignaciouy:
Now when the espeak plugin is not installed there will be an
error (logging.error) in the log. Instead of the ugly message
of Gst

Fixes #3345

GCI Task, https://www.google-melange.com/gci/task/view/google/gci2014/5789378047639552

### DIFF
--- a/src/sugar3/speech.py
+++ b/src/sugar3/speech.py
@@ -18,6 +18,7 @@ import os
 import logging
 from gettext import gettext as _
 
+from gi.repository.GLib import GError
 from gi.repository import Gio
 from gi.repository import Gst
 from gi.repository import Gtk
@@ -277,8 +278,13 @@ class _GstSpeechPlayer(GObject.GObject):
         if self._pipeline is not None:
             self.stop_sound_device()
             del self._pipeline
+            self._pipeline = None
 
-        self._pipeline = Gst.parse_launch(command)
+        try:
+            self._pipeline = Gst.parse_launch(command)
+        except GError:
+            logging.error('The speech plugin is not installed in the system.')
+            return
 
         bus = self._pipeline.get_bus()
         bus.add_signal_watch()
@@ -297,6 +303,9 @@ class _GstSpeechPlayer(GObject.GObject):
             return
 
         self.make_pipeline('espeak name=espeak ! autoaudiosink')
+        if self._pipeline is None:
+            return
+
         src = self._pipeline.get_by_name('espeak')
 
         src.props.text = text


src/sugar3/speech.py
```diff
@@ -278,7 +278,12 @@ def make_pipeline(self, command):
             self.stop_sound_device()
             del self._pipeline
 
-        self._pipeline = Gst.parse_launch(command)
+        try:
+            self._pipeline = Gst.parse_launch(command)
+        except:
```
SAMdroid-apps: Does it have to be a catch all?  Usually a more specific error is better.

ignaciouy: @SAMdroid-apps what about:

except Exception as error:
\t logging.error('The spech plugin can't be found, %s' %  error)

SAMdroid-apps: @ignaciouy, I was talking about something like:

    except ErrorType:
        ...

You should not use catch all except clauses, as you can catch unrelated errors.

godiard: @ignaciouy, can you uninstall gstreamer espeak plugin and check what error do you receive?
Check on _init_voices method too.
I think we can assume Gstreamer is installed and working for this case. If not we can't import Gst and can't do Gst.init() neither.

ignaciouy: @godiard instead of uninstall, I tried to change the plugin name, to random thing, the error I got is GError. btw, working on the Gst.Init patch.

godiard: Not sure if you will get the same error by rename than by uninstall. Please
check if you can.

On Mon, Jan 5, 2015 at 1:53 PM, Ignacio Rodríguez <notifications@github.com>
wrote:

> In src/sugar3/speech.py
> <https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/184#discussion-diff-22472585>
> :
>
> > @@ -278,7 +278,12 @@ def make_pipeline(self, command):
> >              self.stop_sound_device()
> >              del self._pipeline
> >
> > -        self._pipeline = Gst.parse_launch(command)
> > +        try:
> > +            self._pipeline = Gst.parse_launch(command)
> > +        except:
>
> @godiard <https://github.com/godiard> instead of uninstall, I tried to
> change the plugin name, to random thing, the error I got is GError. btw,
> working on the Gst.Init patch.
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/184/files#r22472585>
> .
>



-- 
Gonzalo Odiard

ignaciouy: ​
I haven't installed espeak module on my ubuntu 15.04, that is what I got ^^

Ignacio Rodríguez
nachoel01@gmail.com
SugarLabs at Facebook
<https://www.facebook.com/pages/SugarLabs/187845102582>

2015-01-05 15:04 GMT-02:00 Gonzalo Odiard <notifications@github.com>:

PR #187, opened 2015-01-05

SAMdroid-apps:
Replaces #185 

GCI Ticket:  http://www.google-melange.com/gci/task/view/google/gci2014/5810064153640960

@godiard Please see the comment about application mimes, line 112

### DIFF
--- a/src/sugar3/mime.py
+++ b/src/sugar3/mime.py
@@ -21,13 +21,16 @@
 STABLE.
 """
 
+import re
 import os
 import logging
 import gettext
+from subprocess import check_output
 
 from gi.repository import GLib
 from gi.repository import GdkPixbuf
 from gi.repository import Gio
+from gi.repository import GObject
 
 _ = lambda msg: gettext.dgettext('sugar-toolkit-gtk3', msg)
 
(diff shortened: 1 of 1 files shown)


quozl:
What is the impact on startup time?

SAMdroid-apps:
Very little.  I didn't notice anything.

The 1st patch took about 0.6s to run when I tested it.  And this is only
called when a a/v mime is possibly needed.

quozl:
0.6 seconds sounds okay.  But what hardware did you test on?  Were the filesystem caches empty or populated?

src/sugar3/mime.py
```diff
@@ -94,6 +93,33 @@ def _get_supported_image_mime_types():
 }]
 
 
+def _set_av_mimes():
+    global _generic_types
```
SAMdroid-apps: Oops, I don't need this global, I will re push

SAMdroid-apps:
Ok, so I did some more tests.  On an XO4 with on no file system caches, the test took 1.442s.  On my laptop (Core 2 Duo Mobile @ 2GHz) with file caches cleared, it took 1.27s at 50% cpu.  What do you think?

SAMdroid-apps:
Replacing for some caching :musical_note:

PR #188, opened 2015-01-06
Get audio and video mimes from GStreamer, fixes #3826

SAMdroid-apps:
Now with caching!

Replaces #187

GCI Ticket:  http://www.google-melange.com/gci/task/view/google/gci2014/5810064153640960

PR #190, opened 2015-01-07

SAMdroid-apps:
Replaces #188

GCI Ticket: http://www.google-melange.com/gci/task/view/google/gci2014/5810064153640960

### DIFF
--- a/src/sugar3/mime.py
+++ b/src/sugar3/mime.py
@@ -21,13 +21,19 @@
 STABLE.
 """
 
+import re
 import os
+import json
 import logging
 import gettext
+from subprocess import check_output
 
 from gi.repository import GLib
 from gi.repository import GdkPixbuf
 from gi.repository import Gio
+from gi.repository import GObject
+
+from sugar3.env import get_profile_path
 
 _ = lambda msg: gettext.dgettext('sugar-toolkit-gtk3', msg)
 
@@ -37,6 +43,11 @@ GENERIC_TYPE_AUDIO = 'Audio'
 GENERIC_TYPE_VIDEO = 'Video'
 GENERIC_TYPE_LINK = 'Link'
 GENERIC_TYPE_BUNDLE = 'Bundle'
+
+_AUDIO_RE = re.compile('audio/[A-Za-z0-9\-]+$', re.M)
+_VIDEO_RE = re.compile('video/[A-Za-z0-9\-]+$', re.M)
+_APPLICATION_RE = re.compile('application/[A-Za-z0-9\-]+$', re.M)
+_GST_MIMES_CACHE = get_profile_path('gst_mimes.json')
 
 
 def _get_supported_image_mime_types():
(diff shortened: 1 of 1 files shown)


SAMdroid-apps:
@godiard fixed

quozl:
_set_av_mimes() does not explicitly return a boolean, so the default is False, which will make it run once.  Perhaps a return False would make that clearer.

This will still result in a short hang during startup, since the idle add task will block the UI while waiting for several subprocesses to complete.

Repeating and editing my comment from the previous pull request ... since much of the delay is in synchronous subprocesses, you may instead run those asynchronously and handle the output as it arrives, without blocking the UI.

SAMdroid-apps:
Hi @quozl,

I just re-pushed this patch to add the `return False`

Re: Multithreading, subprocess...  I think we are completely over thinking this issue.  It is a 2 second lag, that the user will experience **once**.  **ONCE**.  Please consider that a good portion of our users are on single core processors, and will see no difference (maybe idle-add will have less overhead of not having to start another python instance).

Re: we should use tickets more... I don't get enough emails to remember that we have a bug tracker (hint: why dosn't tarc email?).  I also feel that tarc is not as sleek and usable as GitHub.  Maybe you should lead by example, email the mailing list or create a ticket.  Hopefully your bug ticket won't be forgotten in the static-feeling, talking-to-a-brick-wall site that bugs.sl.o is (hint: I always feels like your talking to yourself on tarc, there is no real form of conversation).

godiard:
Ok. I think the change is ok to merge.
Can you fix these:
```
[gonzalo@localhost sugar-toolkit-gtk3]$ pep8 src/sugar3/mime.py
src/sugar3/mime.py:121:69: E502 the backslash is redundant between brackets
src/sugar3/mime.py:126:1: W293 blank line contains whitespace
src/sugar3/mime.py:127:5: E303 too many blank lines (2)
```

About the comments here or in the trac, I will reply in the mailing list, because I think have value discuss it more openly.

quozl:
@Sam, first impressions of performance are important.  It might take two seconds for you, but on an XO-1 ... you have no idea how long it takes.  It isn't only CPU performance, there's the filesystem performance as well.  If you provide a patch relative to Sugar 0.98 then I can test it for you.

I'm happy to accept Gonzalo's merge, but if I'm making XO-1 builds for a school I'll probably reverse the patch, or not recommend 0.104 until the performance impact is characterised.

godiard:
@quozl, could be possible create the cache file when the image is build?
Another issue is, the first boot ask to the user the color/age/gender, probably, if the operation is run in a separated thread, should not be so bad...

quozl:
@godiard, agreed.  Separate thread, or asynchronous subprocess Popen, is the same result in the end.

sam:
Wrong "@Sam".

quozl:
Apologies.  Now I know scope is global.

SAMdroid-apps:
Just added the multiprocessing... pushing...

PR #193, opened 2015-01-08
Log error when the espeak plugin is not installed.

ignaciouy:
Now when the espeak plugin is not installed there will be an
error (logging.error) in the log. Instead of the ugly message
of Gst
Also, if Gst isnt installed it debug a error too.

Fixes #3345

Requires https://github.com/sugarlabs/sugar/pull/471


godiard:
merged

PR #194, opened 2015-01-11

samdroid-apps:
Replaces #190

This uses multiprocessing to do the computation in the background.  (Not multithreading because of the GIL...)

### DIFF
--- a/src/sugar3/mime.py
+++ b/src/sugar3/mime.py
@@ -21,13 +21,19 @@
 STABLE.
 """
 
+import re
 import os
+import json
 import logging
 import gettext
+from subprocess import check_output
+from multiprocessing import Process
 
 from gi.repository import GLib
 from gi.repository import GdkPixbuf
 from gi.repository import Gio
+
+from sugar3.env import get_profile_path
 
 _ = lambda msg: gettext.dgettext('sugar-toolkit-gtk3', msg)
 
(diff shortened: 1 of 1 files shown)


SAMdroid-apps:
Requires https://github.com/sugarlabs/sugar/pull/472

SAMdroid-apps:
Just force pushed @godiard, now passes flake 8

quozl:
@godiard, this is first use of "import multiprocessing".  Would it be better to implement using subprocess.Popen and GLib.source_add_*?

SAMdroid-apps:
@quozl Let's try to not make the whole code more complex than it should be.  Multiprocessing is simple and pythonic and does the job fine.  It works in contexts where there is no GSomething main loop.

quozl:
Is this called where there is no main loop?

SAMdroid-apps:
It could be... mimes may be used for scripts and stuff

quozl:
Yes, it could be, everything could be ... but is it actually called by anything that isn't using an event main loop?  That is to say, is there a hard requirement for the code to avoid dependency on the event main loop?

SAMdroid-apps:
@quozl Are you ok with this patch?  We need to make a decision today as the feature freeze is today (in oz time anyway)!

SAMdroid-apps:
Other freeze sorry, the feature freeze is way past

quozl:
Testing on XO-1, standby.

PR #196, opened 2015-01-14
Raise the palette window when its mapped

ignaciouy:
This patch raise the palette window when its mapped
Its use: Gtk.Window.present [1] the user will
be able to saw the palette now when your apply the
test case of #4463

[1] https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-present
Fixes #4463


tchx84:
Merged.

PR #197, opened 2015-01-19
autopep8, flake8 to sugar-toolkit-gtk3 code

ignaciouy:
This fixes the 'osbuild check sugar-toolkit-gtk3' error.
Pre-patch: (check-modules.log) http://fpaste.org/171260/
After-patch (check-modules.log) http://fpaste.org/171261/

### DIFF
--- a/examples/animator.py
+++ b/examples/animator.py
@@ -6,6 +6,7 @@ from sugar3.graphics import style
 
 
 class _Animation(animator.Animation):
+
     def __init__(self, icon, start_size, end_size):
         animator.Animation.__init__(self, 0.0, 1.0)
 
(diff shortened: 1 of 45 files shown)


godiard:
I think this is a huge patch mixing different things.
Would be good have one patch solving the unneeded imports, another changing the deprecated exception handling,  and maybe another with the new lines needed (this is a new rule from pep8).
Not sure why is needed change 0644 to 0o644.   
Probably is safer make the pep8 fixes in the new cycle.

godiard:
Apparently,  the change from 0644 to 0o644 is needed to use python3

https://groups.google.com/forum/#!topic/mezzanine-users/nk_ILlwZc1o

tchx84:
Thanks Ignacio for the noticing this. Sadly, is too late for all these changes at this stage of the release. We can do a major cleanup in the next cycle.

PR #198, opened 2015-03-29
Fix the badge location for icons

SAMdroid-apps:
This is visibly an issue where the badge goes outside the bounds of
the icons.  For example, before this patch the wifi icons look like
this:

[Gyazo](http://gyazo.com/5501940192d95d9df59323d0ddc45af6)

Whereas, after the patch:

[Gyazo](http://gyazo.com/f5780d29bf18f82b1ea3a2844297afce)

### DIFF
--- a/src/sugar3/graphics/icon.py
+++ b/src/sugar3/graphics/icon.py
@@ -202,8 +202,8 @@ class _IconBuffer(object):
             return info
 
         info.size = int(_BADGE_SIZE * icon_width)
-        info.attach_x = int(icon_info.attach_x * icon_width - info.size / 2)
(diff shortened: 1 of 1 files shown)


godiard:
Is possible something else changed? Looking "git blame" the code in _get_badge_info didn't chnaged in years. Also, the badge in the neighborhood wifi spot icons is supposed to appear at the bottom right.
See https://github.com/godiard/help-activity/blob/master/images/Neighborhood_main_annotated.png
Do you know when this bug started to appear?

SAMdroid-apps:
Ah, that makes sense now.

I think this bug came along about f20 time, maybe f21.  I'm not really sure, but it has been around for a while.

I will investigate again tomorrow.

PR #199, opened 2015-03-31
Use SVG metadata for icon badge attachment points

SAMdroid-apps:
Replaces #198

> `gtk_icon_info_get_attach_points` has been deprecated since version 3.14 and should not be used in newly-written code.
>
> Attachment points are deprecated

From https://developer.gnome.org/gtk3/stable/GtkIconTheme.html#gtk-icon-info-get-attach-points

This commit uses the new `sugar:attach-point` metadata to get the attachment
point.  GTK has returned (0, 0) since version 3.14, meaning that the network
view WiFi icons (just as an example) have looked a little bit off.

### DIFF
--- a/src/sugar3/graphics/icon.py
+++ b/src/sugar3/graphics/icon.py
@@ -37,6 +37,7 @@ from sugar3.graphics.xocolor import XoColor
 from sugar3.util import LRU
 
 _BADGE_SIZE = 0.45
+_ATTACH_POINTS_RE = re.compile('sugar:attach-points="([0-9,]+)"')
 
 
 class _SVGLoader(object):
@@ -44,7 +45,7 @@ class _SVGLoader(object):
     def __init__(self):
         self._cache = LRU(50)
 
-    def load(self, file_name, entities, cache):
+    def _get_icon(self, file_name, cache):
         if file_name in self._cache:
             icon = self._cache[file_name]
         else:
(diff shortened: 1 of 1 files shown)


SAMdroid-apps:
Requires https://github.com/sugarlabs/sugar-artwork/pull/45 for the SVGs

quozl:
Tested.  Breaks display of WiFi icons on Fedora 18 with Sugar 0.104.  Screenshot: captura pantalla de _malla_

SAMdroid-apps:
Did you update the icon theme?  Can you double check the icon changes flowed through (I had major issues getting the icon changes reflected)?

quozl:
Ok, no, I missed that.  I'll try again tomorrow, hopefully.

godiard:
HI @SAMdroid-apps , @quozl, i have pushed a alternative solution for this problem (https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/200) where we don't need modify the artwork.
I think would be better keep the information for the attach points separated from the svg.
What you think? Can you test and compare?

SAMdroid-apps:
I find having a whole file for 1 little piece of metadata a little weird (personally).

SAMdroid-apps:
@quozl, can you test this with the icons?

quozl:
@godiard, i've no issue with either method of coding the metadata.
@SAMdroid-apps, i've not had time, testing of icon.py file change was a mv, wget, and restart.  testing of icons looks a whole lot harder.  do you have a test method?

godiard:
We merged the alternative solution.

PR #200, opened 2015-04-01
Read icon attach points when is not possible get the info from gtk

godiard:
Gtk has deprecated gtk-icon-get-attach-points on version 3.14
We need that information to place pur badges,
this fix read the information from the .icon file

[1] https://developer.gnome.org/gtk3/stable/GtkIconTheme.html#gtk-icon-info-get-attach-points

### DIFF
--- a/src/sugar3/graphics/icon.py
+++ b/src/sugar3/graphics/icon.py
@@ -24,6 +24,8 @@ STABLE.
 import re
 import math
 import logging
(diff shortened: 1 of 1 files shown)


quozl:
Tested.  On XO-1.5, using Fedora 18 with Sugar 0.104 and GTK 3.6.4.  No issues seen.

PR #201, opened 2015-04-03
Don't break activity startup if activity info have 'class' instead of 'e...

godiard:
...xec'

When we ported sugar to grtk3 we removed he support to the fields
exec and service_name due they were deprecated for a long time [1].
But this is still a problem in deploymnts, and we already added compatibility
for the service_name field.
Here [3] there is a long thred were we discussed this issue.

[1] ae85dd62d42a7d5bea76aec3b941ce35ccc61bf4
[2] 32f97b48c31f60f306854753a3df734f63b798fc
[3] http://lists.sugarlabs.org/archive/sugar-devel/2013-October/045404.html

### DIFF
--- a/src/sugar3/bundle/activitybundle.py
+++ b/src/sugar3/bundle/activitybundle.py
@@ -101,7 +101,6 @@ class ActivityBundle(Bundle):
 
     def __init__(self, path, translated=True):
         Bundle.__init__(self, path)
-        self.activity_class = None
         self.bundle_exec = None
 
         self._name = None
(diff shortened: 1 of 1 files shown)


SAMdroid-apps:
@godiard This patch works a treat.  I will merge it.  Can you just read through your commit message and resend it?

godiard:
@SAMdroid-apps, other than the typo in 'deploymnts', do you see anything more to change?

SAMdroid-apps:
"Grtk3" and the title spills over a bit

PR #202, opened 2015-04-06
Don't break activity startup if .info have 'class' instead of 'exec'

godiard:
When we ported sugar to Gtk3 we removed the support of the fields
exec and service_name due they were deprecated for a long time [1].
But this is still a problem in deployments, and we already added
compatibility for the service_name field.
Here [3] there is a long thred were we discussed this issue.

[1] ae85dd62d42a7d5bea76aec3b941ce35ccc61bf4
[2] 32f97b48c31f60f306854753a3df734f63b798fc
[3] http://lists.sugarlabs.org/archive/sugar-devel/2013-October/045404.html


SAMdroid-apps:
Merged!

PR #209, opened 2015-04-23
Bundle launcher

samdroid-apps:
Required by https://github.com/sugarlabs/sugar/pull/491

## Add ability to launch a bundle

This commit adds the `launch_bundle` function in `sugar3.activity.
activity` which accesses the same function in the shell is over dbus.
This is means the shell process launches the bundle.  Activities
should not launch child processes as per the Rainbow security model
[1].

[1]  http://wiki.laptop.org/go/Rainbow


godiard:
@samdroid-apps, please move all tthe pep8 fixes unrelated to the feature to another pr

PR #210, opened 2015-04-23
Add ability to launch a bundle

samdroid-apps:
This commit adds the `launch_bundle` function in `sugar3.activity.
activity` which accesses the same function in the shell is over dbus.
This is means the shell process launches the bundle.  Activities
should not launch child processes as per the Rainbow security model
[1].

[1]  http://wiki.laptop.org/go/Rainbow

PEP8 Compliant.

Replaces #209 

Required by https://github.com/sugarlabs/sugar/pull/492

### DIFF
--- a/src/sugar3/activity/activity.py
+++ b/src/sugar3/activity/activity.py
@@ -149,6 +149,7 @@ class _ActivitySession(GObject.GObject):
 
 
 class Activity(Window, Gtk.Container):
+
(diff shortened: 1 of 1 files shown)


quozl:
Looking for the corresponding patch to Browse to use the function, where can it be seen?  Have checked https://github.com/sugarlabs/browse-activity/pulls

samdroid-apps:
I have not submitted a patch to browse yet.

PR #212, opened 2015-05-07
Honor ignore on git

godiard:


samdroid-apps:
This sounds like the right solution

godiard:
@samdroid-apps, something like https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/218 ?

PR #213, opened 2015-05-10
Tick based animation

samdroid-apps:
Gtk 3.8 introduces the tick callback [1].  This allows widgets to be
signaled before each frame, simmilar to requestAnimationFrame in the
browser.

This patch adds an optional widget argument to the Animator class so
tick based animation can be used.  This is much more efficent than
using timeouts, as we get a more appropriate frame rate for the user.

[1] https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-add-tick-callback

### DIFF
--- a/src/sugar3/graphics/animator.py
+++ b/src/sugar3/graphics/animator.py
@@ -34,12 +34,13 @@ class Animator(GObject.GObject):
         'completed': (GObject.SignalFlags.RUN_FIRST, None, ([])),
     }
 
-    def __init__(self, duration, fps=20, easing=EASE_OUT_EXPO):
+    def __init__(self, duration, fps=20, easing=EASE_OUT_EXPO, widget=None):
         GObject.GObject.__init__(self)
         self._animations = []
         self._duration = duration
         self._interval = 1.0 / fps
         self._easing = easing
+        self._widget = widget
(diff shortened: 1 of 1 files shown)


godiard:
@samdroid-apps, could you explain what test you did and what are the results?
Would be good have more information. Also, we need test, specially with old hardware like 
the XO-1

samdroid-apps:
To test, just observe how much more fps the animations are!  They should look silky smooth on recent (eg. 2007 laptop) hardware.

On slower devices (eg. XO4) the change is less visible, because it is based on the devices performance at an given time.

On even slower devices (eg. XO1) they may be less fps, but less fps is way we seem to want on XO1... I can't test though. (CC @quozl)

BTW:  the tick events are introduced in Fedora 19 - so there will be no change on f18 based builds

quozl:
Happy to test, but I need a test plan.  It isn't clear what I have to do to test it.  As this involves another pull request, I'd hope that a ticket at bugs.sugarlabs.org might be the best place to organise this.

samdroid-apps:
Feature page:  http://wiki.sugarlabs.org/go/Features/Tick_based_animation

samdroid-apps:
I will rebase this and retest this

PR #214, opened 2015-05-12
2015 05 silence power

quozl:
Silence some errors associated with the power manager, and remove unnecessary filesystem accesses.

Tested on XO-1 and commodity hardware with Sugar 0.104 and Clock-17.


samdroid-apps:
Merged!

PR #216, opened 2015-05-13
Cell renderer icon optimizations

godiard:
These are simple optimizations, are the first part of the treeview optimization described 
in the feature [1].
* avoid execute slow methods while the icon is rendered
* remove a unneeded operation
* cache size offset calculation
Are isolated from other changes and do not change the API.

[1] http://wiki.sugarlabs.org/go/Features/Optimize_TreeViews

### DIFF
--- a/src/sugar3/graphics/icon.py
+++ b/src/sugar3/graphics/icon.py
@@ -983,7 +983,14 @@ class CellRendererIcon(Gtk.CellRenderer):
         context.save()
         context.add_class("sugar-icon-cell")
 
-        pointer_inside = self._point_in_cell_renderer(widget)
+        def is_pointer_inside():
+            # widget is the treeview
+            x, y = widget.get_pointer()
+            x, y = widget.convert_widget_to_bin_window_coords(x, y)
+            return ((cell_area.x <= x <= cell_area.x + cell_area.width)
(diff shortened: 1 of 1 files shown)


quozl:
Reviewed.  Applied on filesystem of an XO-1 13.2.4 and tested.  The activities list and journal scroll much faster.  You have my vote.

samdroid-apps:
Interesting bug:

When you scroll down, then click a star icon you are scrolled *almost* back to the very top.

godiard:
@samdroid-apps, yes it's strange, but is not related with this changes (I tested on master branch too). Can you fill a bug?

samdroid-apps:
Ok

samdroid-apps:
Created a new ticket [#4852 Journal star icon scrolls to top ](http://bugs.sugarlabs.org/ticket/4852)

samdroid-apps:
Thanks for merging that for me!

Can we have a consensus on if we `git rebase` when we merge?  My vote is that we should (better history in git), but what do you think @godiard or @tchx84?

(Now I understand what it does, sorry about not understanding at the meeting @tchx84 :smile:.  I legit did not know the difference)

godiard:
@samdroid-apps, I think rebase is more clean. But would be better discuss this topic in sugar-devel mailing list

PR #218, opened 2015-05-14
Honor ignore on git try2

godiard:


samdroid-apps:
Merged!

samdroid-apps:
I will deploy this patch to my buildbot docker environment so all new builds should have this patch applied.

PR #219, opened 2015-05-14
Treeview perf2 try2

godiard:
These are two optimizations:

Implement a scrolling detector, (similar to slow mouse detector), that detector can be attached to treeviews, and the cellrendericons are notified when the treeview is scrolling and skip some expensive calculations. The effect is more evident in touch devices with dynamic scroll.
Implement a TreeViewInvoker removing need have multiple CellRendererInvokers. The CellRenderers are not GtkWidgets, then do not have signals needed by the invokers, making the implementation of CellRenderInvoker complicate and making every invoker attach to the same mouse events.
Changes in Sugar are needed to use these improvements, but compatibility with old code is maintained.


godiard:
Replaces https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/217

PR #221, opened 2015-05-17
Fix GtkMenu palette positioning in Gtk 3.16

samdroid-apps:
It fixes it most of the time :smile:

Still no progress on the anti-seizure front.

### DIFF
--- a/src/sugar3/graphics/palettewindow.py
+++ b/src/sugar3/graphics/palettewindow.py
@@ -113,7 +113,7 @@ class _PaletteMenuWidget(Gtk.Menu):
     def set_invoker(self, invoker):
         pass
 
-    def _position(self, widget, data):
+    def _position(self, widget, *data):
         return self._popup_position[0], self._popup_position[1], False
 
     def popup(self, invoker):


godiard:
@samdroid-apps, I am worried of merge this change without be able to test does not break in other environments :/

There are a error message? What changed in data?

samdroid-apps:
@godiard There is a trace back without the patch, because there are too many arguments.

It appears to be that the x and y arguments have changed from "out" arguments [1] to "inout" arguments [2].  This probably means that gir is passing the arguments to the functions now.

We don't need the new arguments and we never needed the user data argument, so we can just use the `*` to dump whatever arguments we get in the data list and forget about them!  This patch will not break compatibility.

[1]  https://developer.gnome.org/gtk3/3.14/GtkMenu.html#GtkMenuPositionFunc
[2]  https://developer.gnome.org/gtk3/stable/GtkMenu.html#GtkMenuPositionFunc

samdroid-apps:
Could we merge this patch before the release?

Otherwise the clipboard palette will be super stuffed up in Fedora 23.

It will not break other environments, see the comment above.

godiard:
Merged. It's low risk and fixes a error.
Would be good if commit header was more explicit bot what is solving,
right now, is not a good registry for the future. (This https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/221#issuecomment-106940548 is a good example of the information that would be nice have in the header)

PR #222, opened 2015-05-19
Add disable-update-mimedb option to bundlebuilder

tchx84:
(NOT READY FOR MERGE YET)

Historically, activities packaging files for debian have been
using bunblebuilder during package creation steps, via "./setup.py
install".

In this scenario, running ActivityBundle.install_mime_type  creates
unncesserary files that can clash with other packages and also create
broken links.

Adds a way to prevent this by passing "--disable-update-mimedb" as an
argument to setup.py. An identical option is present to sugar shell
installation prodecure.

With this option, packaging files can keep using bundlebuilder and
manually add and install these custom mime types, in later packaging
stages, ie., postinst.

This approach maintains backwards compatibility, so current programs
using bundlebuilder will not be affected.

### DIFF
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -253,7 +253,7 @@ class Installer(Packager):
         Packager.__init__(self, builder.config)
         self.builder = builder
 
-    def install(self, prefix):
+    def install(self, prefix, disable_update_mimedb=False):
         self.builder.build()
 
         activity_path = os.path.join(prefix, 'share', 'sugar', 'activities',
@@ -285,7 +285,8 @@ class Installer(Packager):
 
             shutil.copy(source, dest)
 
-        self.config.bundle.install_mime_type(self.config.source_dir)
+        if disable_update_mimedb is False:
+            self.config.bundle.install_mime_type(self.config.source_dir)
 
 
 def cmd_check(config, options):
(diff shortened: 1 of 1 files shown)


tchx84:
@quozl @godiard what do you think of this approach?

godiard:
@tchx84, in general I agree. Added a simple stylistic cooment.

tchx84:
@godiard, there are two reasons for having it as "disable_" (at the CLI level at least), the first reason is backwards compatibility, if we use something like "enable-update-mimedb" it would change the semantics and the behavior of setup.py, the second reason why I used "disabled_updated_mimedb" is because sugar shell provides the exact same option already (even with the same name), and it would be preferable to keep it consistent (easier to understand that modules both do the same thing and to follow up its effects on the code flow).

https://github.com/sugarlabs/sugar/blob/master/Makefile.am#L14

quozl:
Thanks for working on this.

The change looks okay to me, but I don't understand it.  In particular, I still don't understand why setup.py when run for two different activities is causing a file of the same name to be created.  Why isn't the duplication resolved instead?

Minor typo in commit message, s/unncesserary/?

tchx84:
@quozl the problems occurs when setup.py is used during activities packages generation process, to populate the package directories and files (both in Debian and Fedora).

This happens because setup.py runs the "update-mime-database" tool which updates the system database, but because there is no "system database" when the package is being created, it creates a completely new database files, which are included in the package, and since every single activity does the same, these database files collide.

tchx84:
@godiard ok, It would work with default=True, will change that. @quozl thanks for the typo warning.

quozl:
@tchx84, okay, I'm not aware of that problem.  Is there a ticket or bug that describes the problem in detail?

PR #223, opened 2015-05-25
Add escape key to palette windows

quozl:
Popdown a palette window in response to escape key.

Restores the use of the keyboard for interaction with Sugar after startup, without having to use mouse to dismiss the buddy icon menu.

Tested on XO-4 with Sugar 0.105.1.


godiard:
merged

PR #224, opened 2015-05-26
TTS in the toolkit need initialize gstreamer - Fixes #4851

godiard:
If is not initialized, the check for the espeak plugin will fail.
This patch also check for the espeak plugin at the beginnig
avoiding check every time the function enabled() is called.


tchx84:
Merged.

PR #225, opened 2015-05-28
Add disable-update-mimedb option to bundlebuilder

tchx84:
Historically, distro packagers have been using bunblebuilder, via
"setup.py install", to populate packages directories and files.

The install procedure uses ActivityBundle.install_mime_type to install
custom mime types, by creating symlinks and updating the system mime
types database via the update-mime-database tool.

When ActivityBundle.install_mime_type is executed during packages
creation, the symlinks end-up broken in the final package and the use
of update-mime-database populates the final packages with a new copy
of the database files which could clash with the system copy.

This patch adds a new option, called --disable-update-mimedb to setup.py
install to skip the execution of ActivityBundle.install_mime_type and
avoid the issues mentioned above.

Activity packagers should use this new option. The creation of symlinks
and the execution of update-mime-database should be done in post install
packaging steps.

To test in sugar-build:

1. Make sure there is no mime database generated
```
$ cd sugar-build/home/default/data;
$ rm -rf mime;
```

2. Using the shell install an activity with and without the --disable-update-mimedb option.
```
$ ./osbuild shell
$ cd ./read
$ ./setup.py install # and,
$ ./setup.py install --disable-update-mimedb 
```

### DIFF
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -253,7 +253,7 @@ class Installer(Packager):
         Packager.__init__(self, builder.config)
         self.builder = builder
 
-    def install(self, prefix):
+    def install(self, prefix, update_mimedb=True):
         self.builder.build()
 
         activity_path = os.path.join(prefix, 'share', 'sugar', 'activities',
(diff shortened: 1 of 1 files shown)


tchx84:
@godiard please give it a try!

godiard:
@tchx84, Maybe --skip-install-mime es better given that installing the svg icons is also involved

PR #228, opened 2015-05-29
sugar-activity-web: remove bash specific code

quozl:
Gears activity on Ubuntu with Sugar 0.105.1 shows this in logs:

```
/usr/bin/sugar-activity-web: 20: [: unexpected operator
```

Cause is bash-specific coding in script.


godiard:
Thanks, merged. I didn't have idea that == was bash specific...

PR #229, opened 2015-06-19
Secondary text on palettes should respect the max width - Fixes #4862

godiard:
When a activity have a title very long without spaces,
the text can't be wraped with the default wrap mode (WORD)
Use WRAP_CHAR to solve this issue.

### DIFF
--- a/src/sugar3/graphics/palette.py
+++ b/src/sugar3/graphics/palette.py
@@ -286,6 +286,7 @@ class Palette(PaletteWindow):
                 self._secondary_label.set_line_wrap(True)
                 self._secondary_label.set_ellipsize(
                     style.ELLIPSIZE_MODE_DEFAULT)
+                self._secondary_label.set_wrap_mode(Gtk.WrapMode.WORD_CHAR)
                 self._secondary_label.set_lines(NO_OF_LINES)
                 self._secondary_label.set_justify(Gtk.Justification.FILL)
             else:


godiard:
More information in the ticket http://bugs.sugarlabs.org/ticket/4862

quozl:
Reviewed.

(I see now that this is caused by user naming of journal entries rather than developer naming of activities, which is what I initially thought was meant by the ticket and commit message.   ;-)

samdroid-apps:
Seems to be causing a trace back on Gtk 3.17 (Fedora Rawhide).

```
  File "/usr/lib64/python2.7/site-packages/gi/_propertyhelper.py", line 427, in obj_set_property
    prop.fset(self, value)
  File "/home/sam/sb21/build/out/install/lib/python2.7/site-packages/sugar3/graphics/palette.py", line 289, in set_secondary_text
    self._secondary_label.set_wrap_mode(Gtk.WrapMode.WORD_CHAR)
AttributeError: 'Label' object has no attribute 'set_wrap_mode'
```

It appears to work if I use the `.props.wrap_mode`.  Will send a patch.

godiard:
@samdroid-apps, I found this was not the proper solution.
As the patch was already merged, sent a new pr with the real fix.

https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/232

Tested on F22 and F18

PR #230, opened 2015-06-22
translations: add Guarani to LINGUAS

tchx84:
A group of kids from the OLPC project in Caacupe have translated both
the sugar shell and the sugar gtk3 toolkit to Guarani. They are also in
the process of translating sugar activities [1].

This patch add the gn entry to the LINGUAS file so the gn.po file can be
properly installed, as already is being done with the sugar shell.

[1] http://translate.sugarlabs.org/gn/

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -32,6 +32,7 @@ fi
 fil
 fr
 gl
+gn
 gu
 ha
 he


tchx84:
This PR requires https://github.com/sugarlabs/sugar/pull/534

godiard:
@tchx84, please push both.

tchx84:
Merged.

PR #232, opened 2015-06-23
Properly fix wrap mode on secondary text for toolbars

godiard:
This patch solves a bug introduced on 034706a482deef3db5e8a93b232882af3224016c


tchx84:
Merged with a small pep8 fix.

PR #233, opened 2015-06-27
Use CSS for GtkMenu palette header border so the border goes full width

samdroid-apps:
Accompanies sugar artwork patch


samdroid-apps:
Required by:  https://github.com/sugarlabs/sugar-artwork/pull/65

godiard:
Merged alternative version.

PR #235, opened 2015-06-30
Remove _HeaderSeparator

godiard:
The palette using a Gtk.Menu, had a separator displayed using a custom widget,
but that was not well displayed (the line didn't had the palette width
due to margin on the palette) and was broken on Gtk >= 3.16
Instead of use that widget to draw the separator line between at the bottom
of _HeaderItem, as we do with the toolbar buttons when the palette is displayed.


godiard:
Needs https://github.com/sugarlabs/sugar-artwork/pull/66

This is a alternative implementation to solve the problems on GtkMenu palettes reported on https://github.com/sugarlabs/sugar-artwork/pull/65

tchx84:
Merged.

PR #236, opened 2015-07-02
Add get_bundle method to activity.py

godiard:
When a programmer need open one activity from other [1]
usually need display in the user interface information about the activity
to open, like the name or the icon.
This method allow get this information, previously only available on jarabe.

[1] http://wiki.sugarlabs.org/go/Features/Start_activity_from_another_activity

### DIFF
--- a/src/sugar3/activity/activity.py
+++ b/src/sugar3/activity/activity.py
@@ -83,6 +83,7 @@ from sugar3.graphics.alert import Alert
 from sugar3.graphics.icon import Icon
 from sugar3.datastore import datastore
 from sugar3.bundle.activitybundle import get_bundle_instance
+from sugar3.bundle.helpers import bundle_from_dir
 from gi.repository import SugarExt
 
 _ = lambda msg: gettext.dgettext('sugar-toolkit-gtk3', msg)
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Why are we adding the dbus overhead?  The `get_bundle` method is designed
to be runnable outside the jarabe process.

Could we just make this a wrapper without dbus?

godiard:
@samdroid-apps, we need ask to sugar for the bundle installed.

samdroid-apps:
Huh?

godiard:
@samdroid-apps, all the logic used to read the bundles is in the toolkit, but Sugar (and the code is in jarabe) have a registry of the activities installed in sugar. If we don't use dbus, need move all that logic to the toolkit, and the worst part is every activity process need store the registry information.

samdroid-apps:
But it works perfectly fine from outside of the jarabe process right now?

Is there any reason to increase the surface area of the change so late in
the cycle?

quozl:
Yes, good reasons.  Don't want dependency between an activity and sugar.  Hurts everybody in future as sugar changes.  The point of isolation must be the toolkit.

godiard:
@samdroid-apps, I explained the reason in my email to sugar-devel. Use jarabe in the activities is forbidden, because is a internal API and we don't warranty stability.
I know is very late, but in my experience, every time we change the APIs our users have more problems than we expect. Then, I think we need do the release with a usable API.

quozl:
+1

PR #237, opened 2015-07-02
Add get_bundle method to activity.py

godiard:
When a programmer need open one activity from other [1]
usually need display in the user interface information about the activity
to open, like the name or the icon.
This method allow get this information, previously only available on jarabe.

[1] http://wiki.sugarlabs.org/go/Features/Start_activity_from_another_activity


quozl:
Reviewed.

tchx84:
Tested in sugar-build/fedora21.

godiard:
Merged with tch approval.

PR #240, opened 2015-07-10
Deprecate create_jobject parameter on activity.py - Fixes #3450

godiard:
As the ticket describe, create a object in the Journal
is no longer optional. The change do not break activities,
because the optional parameter is not removed, just ignored.

### DIFF
--- a/src/sugar3/activity/activity.py
+++ b/src/sugar3/activity/activity.py
@@ -268,8 +268,9 @@ class Activity(Window, Gtk.Container):
             application
 
         create_jobject -- boolean
-            define if it should create a journal object if we are
-            not resuming
+            DEPRECATED: define if it should create a journal object if we are
+            not resuming. The parameter is ignored, and always  will
+            be created a object in the Journal.
 
         Side effects:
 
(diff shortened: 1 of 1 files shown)


tchx84:
@godiard, I agree this is a easier way out of the problem mentioned in #3450, but is there any particular reason not to go for the solution of covering the unchecked uses of self._jobject?

godiard:
@tchx84, I think is the safe way to avoid this problem. After the ticket was filled we added even more logic using the jobject (Description button). Not save a object in the journal was a option needed when Sugar started a new instance by default, and we not use create_jobject=False in any maintained activity (or will crash) but this is to avoid errors in old activities.

godiard:
You can see part of the uses of jobject on widgets.py https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/src/sugar3/activity/widgets.py

samdroid-apps:
This now conflicts with the updated documentation.  This looks like a good patch though, and nobody seems interested in fixing it.  @godiard can you submit a new version of the patch?

godiard:
@samdroid-apps, sent a new pr https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/257

PR #244, opened 2015-07-25
Show proper state indication for CanvasIcons

samdroid-apps:
Commit a19cf9e changed the logic for handling CanvasIcon clicks.  Therefore, the visual state indicators must be updated to fit the new logic.

Test Case:  Hold the mouse down on a home view icon.  Move the mouse in and out of the icon.  See that the states correspond with what happens if you release the mouse.


godiard:
Merged

PR #247, opened 2015-08-07
Do not calculate palette height on PaletteWindowWidget - Fixes #4792

godiard:
To solve ticket #4673, we calculate the palette size using the size
of all the widgets [2]. That solves the problem on Palettes based on
_PaletteMenuWidget but on _PaletteWindowWidget produce a wrong value
when the secondary_label is not null, and as a consequence
the palette position is wrong.
The solution is do the calculation only for _PaletteMenuWidget based
palettes.

[1] https://bugs.sugarlabs.org/ticket/4673
[2] b9d6b628a98fc17f5e92a344450b6336f1a2b2ce

### DIFF
--- a/src/sugar3/graphics/palettewindow.py
+++ b/src/sugar3/graphics/palettewindow.py
@@ -621,18 +621,19 @@ class PaletteWindow(GObject.GObject):
             return
 
         req = self._widget.size_request()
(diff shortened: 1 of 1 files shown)


godiard:
These palette issues are tricky, please test carefully.

samdroid-apps:
Does not create any visual changes in write activity, the shell or the journal

PR #248, opened 2015-08-07
Docs start

samdroid-apps:
@godiard and I have been discussing working on docs for sugar3.  I'll make a feature page soon to explain this all :smile:

PR #250, opened 2015-08-10
Docs try2

samdroid-apps:
Feature page:  http://wiki.sugarlabs.org/go/Features/Sugar3Docs

Replaces #248 

Now has the makefile included


godiard:
I found many modules (like activity.py) didn't generate docs, due to a error trying to import  datastore.py. I propose this https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/251 to solve the probem.

godiard:
Other than that, I think is a very good start, maybe we can tweak how deep is the index, by example if is possible remove the "Subpackages", "SubModules"  entries and so

godiard:
I think we are ready to go, and merged  the patches. (used cherry-pick to keep a sane order). Running make-doc.sh, I found another issue, but I don't know what would be the best solution: webkit1 and webkit2 collision. Affects only src/sugar3/activity/webkit1.py

samdroid-apps:
Do we still use the webkit1 one?

Don't we now have webkit2 on the xos?

Maybe we could aim to be WK1 free in sugar 1.0/0.108?

godiard:
No, we still need it on Fedora 18. And XOs are still on F18

samdroid-apps:
:(  maybe we could move it to jarabe?  It really is not a public api

samdroid-apps:
WebKit2 is on the OLPC OS 0.106 by the way, but it seems to segfault.  That's a big annoying with browse :)

godiard:
That is the reason we still need WebKit (1)

FGrose:
Please update the wiki link to https://wiki.sugarlabs.org/go/Features/Sugar3_Docs

PR #251, opened 2015-08-10
Allow import datstore.py without glib loop

godiard:
Right now is not possible do "from sugar3.datastore import datastore"
outside of GLib loop, because the import try to connect to a DBus signal,
and throw a exception.

That was not a problem in Sugar, because we always have a Glib loop
on all our activities, but is a problem when we want import the classes
to generate docs with pydoc. Many other classes import datastore,
then this is a blocker.

This patch add a try catch to allow this use case.

### DIFF
--- a/src/sugar3/datastore/datastore.py
+++ b/src/sugar3/datastore/datastore.py
@@ -48,9 +48,12 @@ def _get_data_store():
         _data_store = dbus.Interface(_bus.get_object(DS_DBUS_SERVICE,
                                                      DS_DBUS_PATH),
                                      DS_DBUS_INTERFACE)
-        _data_store.connect_to_signal('Created', __datastore_created_cb)
-        _data_store.connect_to_signal('Deleted', __datastore_deleted_cb)
-        _data_store.connect_to_signal('Updated', __datastore_updated_cb)
+        try:
+            _data_store.connect_to_signal('Created', __datastore_created_cb)
+            _data_store.connect_to_signal('Deleted', __datastore_deleted_cb)
+            _data_store.connect_to_signal('Updated', __datastore_updated_cb)
+        except RuntimeError:
+            logging.error('Error trying to connect to datastore')
 
     return _data_store
 


samdroid-apps:
Looks good but maybe we should log the exception in case an actual error happens connecting to the data store?

godiard:
Yes, is already logging....

quozl:
@samdroid-apps, do you mean log the exception detail rather than hide it?

samdroid-apps:
Yes, that is what I mean @quozl.

If the user getsba runtime error then we should log the traceback

godiard:
The exception is cryptic if you don't understand why is happening:

/home/gonzalo/sugar-build/sugar-toolkit-gtk3/doc/sugar3.datastore.datastore.rst:4: WARNING: autodoc: failed to import module u'sugar3.datastore.datastore'; the following exception was raised:
  File "/usr/lib/python2.7/site-packages/sphinx/ext/autodoc.py", line 335, in import_object
    __import__(self.modname)
  File "/home/gonzalo/sugar-build/sugar-toolkit-gtk3/src/sugar3/datastore/datastore.py", line 75, in <module>
    _get_data_store()
  File "/home/gonzalo/sugar-build/sugar-toolkit-gtk3/src/sugar3/datastore/datastore.py", line 51, in _get_data_store
    _data_store.connect_to_signal('Created', __datastore_created_cb)
  File "/usr/lib64/python2.7/site-packages/dbus/proxies.py", line 544, in connect_to_signal
    dbus_interface, **keywords)
  File "/usr/lib64/python2.7/site-packages/dbus/proxies.py", line 369, in connect_to_signal
    **keywords)
  File "/usr/lib64/python2.7/site-packages/dbus/bus.py", line 148, in add_signal_receiver
    path, **keywords)
  File "/usr/lib64/python2.7/site-packages/dbus/connection.py", line 400, in add_signal_receiver
    self._require_main_loop()
RuntimeError: To make asynchronous calls, receive signals or export objects, D-Bus connections must be attached to a main loop by passing mainloop=... to the constructor or calling dbus.set_default_main_loop(...)

quozl:
What of the exceptions that might occur in context other than the use case?

quozl:
Looks like this was an unforeseen regression introduced in e1d4afe13e3b49f01083a00ca4b83cc4011829d8 which was not intended to bring initialisation forward for any performance reason.

I suggest instead to defer the initialisation to after import, meaning to call _get_data_store on first executable use rather than at import.

samdroid-apps:
Merged alternative patch (#252)

PR #252, opened 2015-08-11
Init the datastore on demand

godiard:
Right now is not possible do "from sugar3.datastore import datastore"
outside of GLib loop, because the import try to connect to a DBus signal,
and throw a exception.

That was not a problem in Sugar, because we always have a Glib loop
on all our activities, but is a problem when we want import the classes
to generate docs with pydoc. Many other classes import datastore,
then this is a blocker.

This patch removes the initialization when imported,
will be done when needed.

### DIFF
--- a/src/sugar3/datastore/datastore.py
+++ b/src/sugar3/datastore/datastore.py
@@ -72,8 +72,6 @@ created = dispatch.Signal()
 deleted = dispatch.Signal()
 updated = dispatch.Signal()
 
-_get_data_store()
-
 
 class DSMetadata(GObject.GObject):
     """A representation of the metadata associated with a DS entry."""


godiard:
This is another alternative to https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/251.
Surprisingly, works.
I really don't understand why this code was written in this way in the first place, but is really old, and is a key piece, then need deep testing.

quozl:
Reviewed.  Looks fine.  Suggest testing in 0.107.

Also reviewed whole file:
 * good; _bus is local,
 * good; no direct use of _data_store global later,
 * risk; propagation of created, deleted, and updated signals from the datastore will be blocked until first call.

Considering this risk, checked existing code; known callers that use the signals; only one;
 * good; src/jarabe/desktop/favoritesview.py:ActivityIcon subscribes to signals in \__init__() but calls _refresh() which calls _get_last_activity_async() which calls datastore.find(),

Callers that do not use the signal but could;
 * src/jarabe/journal/model.py:_get_datastore bypasses the datastore.py API and uses D-Bus.
 * Browse.activity/downloadmanager.py:_create_journal_object bypasses the datastore.py API and uses D-Bus.

No activity source was found that used datastore.{created,updated,deleted}.connect;

So your change looks quite safe.

samdroid-apps:
Thank you for all of the good testing and analysis @quozl.

I will merge this patch.

PR #253, opened 2015-08-11
Improve documentation on activity module

godiard:


godiard:
This is a initial test of the new docs format

godiard:
Reference http://wiki.sugarlabs.org/go/Features/Sugar3Docs

src/sugar3/activity/activity.py
```diff
@@ -19,6 +21,9 @@ class ReadActivity(activity.Activity):
         pass
 
 To get a real, working activity, you will at least have to implement:
+
+.. code-block:: python
+
     __init__(), read_file() and write_file()
```
samdroid-apps: Why do we need to make this a code block?  Could we make these items links and part of the paragraph?  :func:`sugar3.activity.activity.Activity.__init__`

godiard: ok

PR #254, opened 2015-08-25
Improve documentation on activity module

godiard:
Replaces https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/253.

Add links to read_file() and write_file(), but ot to __init__()  because is not documented.
Added __init() in the example code to make it more clear.

Also, formatted better the test at the start describing the module.


samdroid-apps:
Looks great

PR #257, opened 2015-10-05
Deprecate create_jobject parameter on activity.py - Fixes #3450

godiard:
As the ticket describe, create a object in the Journal
is no longer optional. The change do not break activities,
because the optional parameter is not removed, just ignored.


godiard:
Replaces https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/240

samdroid-apps:
Great, merging

godiard:
Thanks for review

PR #258, opened 2015-10-05
Add a widget parameter to the Animator (tick based animations)

samdroid-apps:
Replaces #213 - rebased and few changes to improve reliability

Gtk 3.8 introduces the tick callback [1]. This allows widgets to be
signaled before each frame, simmilar to requestAnimationFrame in the
browser.

This patch adds an optional widget argument to the Animator class so
tick based animation can be used. This is much more efficent than
using timeouts, as we get a more appropriate frame rate for the user.

[1] https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-add-tick-callback

### DIFF
--- a/src/sugar3/graphics/animator.py
+++ b/src/sugar3/graphics/animator.py
@@ -24,15 +24,15 @@ Example:
         from gi.repository import Gtk
         from sugar3.graphics.animator import Animator, Animation
 
-        # Construct a 5 second animator
-        animator = Animator(5)
-
         # Construct a window to animate
         w = Gtk.Window()
         w.connect('destroy', Gtk.main_quit)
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Testing:

1.  Apply this patch and https://github.com/sugarlabs/sugar/pull/572
2.  If running Gtk 3.8+, see smoother animation
3.  If not (eg. Fedora 18), see animations still working

Feature page: http://wiki.sugarlabs.org/go/Features/Tick_based_animation

quozl:
Reviewed.  I've got these animations turned off in OLPC OS packaging of Sugar, so I'm fine with merging.  Looks good.

PR #259, opened 2015-10-10
Position palette based on even coords, fixes #4897

samdroid-apps:
Instead of getting the position of the mouse later in the code,
this commit uses the coords given to sugar by Gtk that say where
the event happened.  This means that if the system is busy, or
the user is fast with the mouse, the palette will pop up where
they originally clicked.

Ticket <https://bugs.sugarlabs.org/ticket/4897>


quozl:
reviewed.

samdroid-apps:
Oops, I'm missing a t from this commit message

PR #260, opened 2015-10-10
Do not give Tube connections to activities, fixes collab

samdroid-apps:
Sugar previously offered a text chan and a tubes chan to each
activity.  This commit changes it to only give a text chan,
because the tubes chan have been removed upstream.

NOTE:  This then gives us 2 options going forward.  Port things to text chans or add new stream tubes api.  Activities that use tubes include image viewer, turtle, browse (why?), pippy, read and write.  There is a patch to port read to stream tubes from 6 years ago in trac:  https://bugs.sugarlabs.org/attachment/ticket/758/tube-stable-api.patch

NOTE2:  @i5o and I tested this on chat activitity.  Then a bunch of kids on XOs joined the chat even though they did not have the patch.  So we have backwards comapt.

List of other bugs found:

[ ] When @i5o came online, he still showed as offline in my buddies view
[ ] File transfers show a gnome save file popup
[ ] People don't show up in the left frame panel
[ ] People didn't huddle around the chat in my neighbourhood view, or @i5o's

### DIFF
--- a/src/sugar3/presence/activity.py
+++ b/src/sugar3/presence/activity.py
@@ -96,7 +96,6 @@ class Activity(GObject.GObject):
         self._account_path = account_path
         self.telepathy_conn = connection
         self.telepathy_text_chan = None
-        self.telepathy_tubes_chan = None
 
         self.room_handle = room_handle
         self._join_command = None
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Replaced by #270

PR #262, opened 2015-10-15
fix alerts documentation

quozl:
- s/Set/Get for a get property method,

- fix two typos.


samdroid-apps:
Looks good, I will merge when I can

PR #263, opened 2015-10-15
remove GSource used by TimeoutAlert

quozl:
For an object of class ```TimeoutAlert```, an OK response occurs on timeout, even if user has clicked a button.

Remove the timeout when the user clicks a button.

A new method ```_response``` is added which overrides the parent method.


samdroid-apps:
Looks good to me

PR #264, opened 2015-10-15
make NotifyAlert inherit from TimeoutAlert

quozl:
Only difference between a ```NotifyAlert``` and a ```TimeoutAlert``` is in the ```__init__```, in particular the ```__timeout``` method is identical to that in ```TimeoutAlert```.

### DIFF
--- a/src/sugar3/graphics/alert.py
+++ b/src/sugar3/graphics/alert.py
@@ -413,7 +413,7 @@ class TimeoutAlert(Alert):
         return True
 
 
-class NotifyAlert(Alert):
+class NotifyAlert(TimeoutAlert):
     """
     Timeout alert with only an "OK" button.  This should be used just for
     notifications and not for user interaction.  The alert will timeout after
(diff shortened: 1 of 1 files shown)


samdroid-apps:
I tested this by applying the patch and using view source (which adds a notify alert at the end).  It did not show the notify alert.  Here is the part of the log that is relevant:

```
  File "/home/saam/sugar-build/build/out/install/lib/python2.7/site-packages/jarabe/view/viewsource.py", line 446, in async_copy_activity_tree
    alert = NotifyAlert(10)
  File "/home/saam/sugar-build/build/out/install/lib/python2.7/site-packages/sugar3/graphics/alert.py", line 460, in __init__
    GLib.timeout_add(1000, self.__timeout)
AttributeError: 'NotifyAlert' object has no attribute '_NotifyAlert__timeout'
```

quozl:
Replaced by https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/266

PR #265, opened 2015-10-15
Position palette based on event coords, fixes #4897

samdroid-apps:
Instead of getting the position of the mouse later in the code,
this commit uses the coords given to sugar by Gtk that say where
the event happened.  This means that if the system is busy, or
the user is fast with the mouse, the palette will pop up where
they originally clicked.

Replaces #259 

Exactly the same as the previous one, except for the commit message.

### DIFF
--- a/src/sugar3/graphics/palettewindow.py
+++ b/src/sugar3/graphics/palettewindow.py
@@ -1008,13 +1008,32 @@ class Invoker(GObject.GObject):
     def notify_mouse_leave(self):
         self.emit('mouse-leave')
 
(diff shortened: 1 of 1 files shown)


quozl:
Not reviewing, no material change since https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/259. ;-)

quozl:
Thanks for the fix, I've cherry-picked it for an OLPC fork.

tchx84:
Tested  and merged.

PR #266, opened 2015-10-16
move guts of TimeoutAlert and NotifyAlert to a private class

quozl:
Create a private class for the button, timeout and response.  Use the private class as parent for both ```NotifyAlert``` and ```TimeoutAlert```.

Extends the spurious response fix (cfeabdc) to ```NotifyAlert```.

Also use ```g_timeout_add``` in preference to ```g_timeout_add_seconds```, because the latter is less precise in implementation.  See https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#g-timeout-add-seconds

Tested by changing the Register alert temporarily.

Replaces https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/264

### DIFF
--- a/src/sugar3/graphics/alert.py
+++ b/src/sugar3/graphics/alert.py
@@ -346,9 +346,35 @@ class _TimeoutIcon(Gtk.Alignment):
         self._text.set_markup('<b>%s</b>' % GLib.markup_escape_text(str(text)))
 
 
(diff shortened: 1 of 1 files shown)


quozl:
I've cherry-picked this for an OLPC fork.

PR #267, opened 2015-10-17
Collab wrapper

samdroid-apps:
WORK IN PROGRESS, FOR REVIEW ONLY

```
commit 86eaf5f45724c63933b8aa6a370028867c3705d7
Author: Sam Parkinson <sam@sam.today>
Date:   Sat Oct 17 20:59:14 2015 +1100

    Add generated documentation to the gitignore

commit f638415e423551aa6ebab7067031a965996d68df
Author: Sam Parkinson <sam@sam.today>
Date:   Sat Oct 17 20:57:04 2015 +1100

    Add a basic, high level abstraction to the collab system
    
    This commit add a high level collab abstraction.  The abstraction
    runs over a text channel, so it works with the telepathy changes.
    
    The abstraction automatically sets up new buddies by sending them
    the state from the leader.  The abstraction allows activities to
    post json encodeable objects to the other users.
```

You need to enable this as the same time as the other collab patch #260 if you have a new telepathy version (after fc20 I think).

This doesn't use any extra encoding other than json (no base64), IDK why there was even a "slash safe" thing in the chat activity.  Test case:  add a bibliography item that uses a slash.

Use bibliography activity to test, here is [the patch to see how simple it is to use the api](https://github.com/samdroid-apps/bibliography-activity/compare/collab?expand=1).  Branch:  https://github.com/samdroid-apps/bibliography-activity/tree/collab

Please read the docs for the new api to see how it can be used.

I still need to look into adding an api to send files between the buddies.  This is required by read activity for example.  However, hopefully I can use something that requires less code than HTTP over TUBES.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,7 @@ libtool
 ltmain.sh
 missing
 compile
+
+/doc/sugar3.rst
+/doc/sugar3.*.rst
+/doc/_build
--- a/src/sugar3/presence/Makefile.am
+++ b/src/sugar3/presence/Makefile.am
@@ -6,5 +6,6 @@ sugar_PYTHON =			\
 	connectionmanager.py	\
 	sugartubeconn.py	\
 	tubeconn.py		\
+	wrapper.py		\
 	presenceservice.py
 
--- a/src/sugar3/presence/wrapper.py
+++ b/src/sugar3/presence/wrapper.py
@@ -1,0 +1,389 @@
+# Copyright (C) 2015 Walter Bender
+# Copyright (C) 2015 Sam Parkinson
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this library; if not, write to the Free Software
+# Foundation, 51 Franklin Street, Suite 500 Boston, MA 02110-1335 USA
+
+'''
+The wrapper module provides an abstraction over the sugar
+collaboration system.
+
+Using CollabWrapper
+-------------------
+1. Implement the `get_data` and `set_data` methods in your activity
+   class::
+
+    def get_data(self):
+        # return plain python objects - things that can be encoded
+        # using the json module
+        return dict(
+            text=self._entry.get_text()
+        )
+
+    def set_data(self, data):
+        # data will be the same object returned by get_data
+        self._entry.set_text(data.get('text'))
+
+2. Make your CollabWrapper instance::
+
+    def __init__(self, handle):
+        sugar3.activity.activity.Activity.__init__(self, handle)
+        self._collab = CollabWrapper(self)
+        self._collab.connect('message', self.__message_cb)
+
+        # setup your activity
+
+        self._collab.setup()
+
+3. Post any changes to the CollabWrapper.  The changes will be sent to
+   other users if any are connected::
+
+    def __entry_changed_cb(self, *args):
+        self._collab.post(dict(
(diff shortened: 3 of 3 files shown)


samdroid-apps:
Hum, found a werid bug where I get an invite to a chat activity if I quit on the owners laptop.

1.  Open bibliography on Bill's laptop
2.  Bill invites Steve to bibliography
3.  Steve accepts the invite
4.  Bill stops the activity
5.  Steve now gets an invite to a chat activity with Steve's colours
6.  Steve posts a message to that chat
7.  Bill now gets an invite to a chat activity, with Bill's colours
8.  Bill and Steve can now chat to each other in the new chat activity

It's really weird.  Maybe it is sugar trying to send what it thinks are non-sugar text channels to chat activity - since that it our chat client.

Tested where Bill was an XO on 14.1 and Steve was a laptop on fc24 rawhide in sugar-build.  Tested over telepathy salut

jvonau:
Just a question as this patch is for salut, can you populate control-panel -> network -> Collaboration -> Server with: jabber.sugarlabs.org and still see the other users in the Network View like SoaS currently does with this patch applied?

samdroid-apps:
@jvonau that code that deals with the accounts in all fine.  There isn't any reason that fixing collaboration will result in touching that code.

jvonau:
I was just concerned with changes to the collaboration stack to fix salut that gabble might become broken, that is why I asked if you tested gabble with a jabber server in place. Is the plan to continue to support both methods? If not then the XSCE project can drop support for ejabberd, used to support sugar's gabble clients when registered with a schoolserver.

samdroid-apps:
Hi Jerry,

From memory, on fc22+ neither gabble or salut work.  That is because we are
using depreciated apis (depreciated since 2009 kind of time) that have been
finally removed recently (the Tubes api).

By using apis that are not depreciated, eg. text or file transfer channels,
we will get better compatibility with gabble, salut and any other account
manager that you want.

Telepathy is all about having the same api for whatever backend technology
you want.  Telepathy basically abstracts away xmpp (gabble) and local link
xmpp (salut) differences.

quozl:
I'd like to see XSCE in the feature page test plan too, not just jabber.sugarlabs.org.

samdroid-apps:
Well you can put it there if you want.  If it has a legit jabber server,
its not gonna make a difference.  I don't even change the accounts code in
any way at all.

quozl:
Good.  I hope you can see how many downstreams are potentially affected and we'd all love to be involved, hence a test plan.  :grin:

jvonau:
Just want to point out F23 SoaS with sugar .106 gabble works out of the box
using  jabber.sugarlabs.org. I have not tested if registration to a
schoolserver still is available and/or still works yet.

samdroid-apps:
Does collaborating in activities work in SoaS 0.106, or does only the mesh buddies view work?

quozl:
@samdroid-apps, I'm confused.  When you say mesh, I think IEEE 802.11s on XO-1, which would be *salut*, not *gabble*.  Or if ad-hoc, then *salut*.  Or maybe you mean the [neighbourhood view](https://help.sugarlabs.org/en/neighborhood_view.html)?

@jvonau, thanks, that's interesting, it means Fedora 23 SoaS has a version of *telepathy-gabble* that works properly with Sugar 0.106, so perhaps no change is needed.  Koji tells me 0.18.2.  What functions of telepathy did you test?  I wish we had a test plan for all the combinations.

jvonau:
I just booted the iso in a vm, noting that someone had browse shared on
j.sl.o. I can share an activity but the "neighbourhood view" doesn't show
on my end but I can't recall if it should or not. I have not connected two
clients to see if others can see the first shared activity. I should really
arrange public ejabberd servers with(out) the olpc patches applied for
testing with, but I lack the web-hosting for that.

quozl:
@jvonau, thanks.  I guess we'll all have to work toward a test plan.  I don't know if jabber.sugarlabs.org has ejabberd OLPC patches or not, but I presume it does, or they are no longer needed.  Either would be useful data.

jvonau:
Don't think jabber is running ejabberd, think it's something else
porosity(can't recall the spelling..) maybe?

samdroid-apps:
Replaced by https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/270

PR #268, opened 2015-10-31
Specify library version prior to import, fixes #4900

samdroid-apps:
Gi has been emitting warnings into the log to do this for some
time now. Not doing this has also caused bugs, such as #4900.

Also see https://github.com/sugarlabs/sugar/pull/600


quozl:
Reviewed.  It's going to be fun when new versions of these GI objects begin to appear.  :grin:

PR #271, opened 2015-11-23
allow start on full disk

quozl:
When disk is full, Sugar will not start, because the backup log directory cannot be created.

X session logs in tmpfs show this exception:

```
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/usr/lib/python2.7/site-packages/jarabe/main.py", line 20, in <module>
    logger.cleanup()
  File "/usr/lib/python2.7/site-packages/sugar3/logger.py", line 113, in cleanup
    os.mkdir(backup_dir)
OSError: [Errno 28] No space left on device:
    '/home/user/.sugar/default/logs/1448258018'
```

Fix is to ignore the exception.

We've been here before!  Was previously fixed in 2010:

 - https://github.com/sugarlabs/sugar/commit/4cde481
 - https://bugs.sugarlabs.org/ticket/1720
 - http://dev.laptop.org/ticket/9623

But new regression introduced in 2012 during reorganisation of code by Daniel Narvaez; the exception was no longer handled:

 - https://github.com/sugarlabs/sugar/commit/0e45f9d
 - https://github.com/sugarlabs/sugar/commit/19db9c5

Reported-by: Nathan Riddle <nathanr333@charter.net>
Tested-by: James Cameron <quozl@laptop.org>

### DIFF
--- a/src/sugar3/logger.py
+++ b/src/sugar3/logger.py
@@ -105,16 +105,21 @@ def cleanup():
                 os.remove(os.path.join(root, f))
             os.rmdir(root)
         except OSError, e:
-            print "Could not remove old logs filesi %s" % e
+            print "Could not remove old logs files %s" % e
 
     if len(backup_logs) > 0:
         name = str(int(time.time()))
         backup_dir = os.path.join(logs_dir, name)
-        os.mkdir(backup_dir)
-        for log in backup_logs:
-            source_path = os.path.join(logs_dir, log)
-            dest_path = os.path.join(backup_dir, log)
-            os.rename(source_path, dest_path)
+        try:
+            os.mkdir(backup_dir)
+            for log in backup_logs:
+                source_path = os.path.join(logs_dir, log)
+                dest_path = os.path.join(backup_dir, log)
+                os.rename(source_path, dest_path)
+        except OSError, e:
+            # gracefully deal w/ disk full
+            if e.errno != errno.ENOSPC:
+                raise e
 
 
 def start(log_filename=None):


src/sugar3/logger.py
```diff
@@ -105,16 +105,21 @@ def cleanup():
...
         name = str(int(time.time()))
         backup_dir = os.path.join(logs_dir, name)
-        os.mkdir(backup_dir)
-        for log in backup_logs:
-            source_path = os.path.join(logs_dir, log)
-            dest_path = os.path.join(backup_dir, log)
-            os.rename(source_path, dest_path)
+        try:
+            os.mkdir(backup_dir)
+            for log in backup_logs:
```
samdroid-apps: Couldn't we put the renaming outside the try?  Moving a file doesn't use space does it?

src/sugar3/logger.py
```diff
@@ -105,16 +105,21 @@ def cleanup():
...
+                source_path = os.path.join(logs_dir, log)
+                dest_path = os.path.join(backup_dir, log)
+                os.rename(source_path, dest_path)
+        except OSError, e:
```
samdroid-apps: Not a biggie, but if you end up changing this patch please use "as e" as it is the new syntax.

samdroid-apps:
Looks like a good fix.  See comments on the patch

quozl:
Yes, a rename may consume space, especially on jffs2.  But if mkdir fails there's no point in doing rename, because the target directory probably does not exist.  So then yet another try except is needed to protect the rename.

Yes, "as e" is the new syntax, but that's a separate issue.  I would reject a patch that used inconsistent coding style while fixing a bug.  There are 17 other places in sugar-toolkit-gtk3 and 57 such in sugar.  You might want to submit a patch for that later?

samdroid-apps:
OK, I'll merge this arvo.

Re as:  if ya want.  It's not a biggie.  We can probably wait until we want
to go python 3 to move it.  I think we were blocked on telepathy python.

PR #270, opened 2015-11-17
Do not give activities telepathy Tubes - they are deprecated

samdroid-apps:
Replaces #260 - does exactly the same thing, just removes more code

Tubes API has been deprecated since 0.17.25 [1], which was probably
around 2009.  This has since been removed from telepathy, breaking
collaberation.

Activities are still given a text channel.  They can also make their
own channels by using the telepathy connection and calling
CreateChannel on the requests interface.

[1]  http://telepathy.freedesktop.org/spec/Channel_Type_Tubes.html

Also, we use RequestChannel rather than the new and not deprecated CreateChannel.  I'll make a patch for that too :smile:

PR #272, opened 2015-12-06
High level wrapper over collaberation (for discussion)

samdroid-apps:
Probs requires https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/270 if your a system that doesn't give out telepathy tubes

This commit introduces the CollabWrapper.  This is a high level
wrapper over the collaberation system.  It uses telepathy text
channels (rather than deprecated tubes used in most activities)
as well as creating file transfer channels to transfer large
messages in a 1-to-1 manner.

For activities, the CollabWrapper provides a simple api.  It syncs
the state when a user joins the activity.  It allows for posting
messages to all other buddies.  It also allows for initiating a
file transfer directly to a buddy, from either a string in memory
or a file on the file system.

Example API usage:  https://github.com/samdroid-apps/bibliography-activity/compare/collab?expand=1

TODO:

* Further test resuming a shared activity
* Discuss if file transfer tubes should be considered backwards
  compatiable, as they popup shell notifications without the
  appropriate patch (https://github.com/sugar/pull/621)
* Test sending a file from a file on the file system
* Test accepting a file to the filesystem

### DIFF
--- a/src/sugar3/presence/filetransfer.py
+++ b/src/sugar3/presence/filetransfer.py
@@ -1,0 +1,358 @@
+# Copyright (C) 2008 Tomeu Vizoso
+# Based on jarabe.model.filetransfer
+# Copyright (C) 2015 Sam Parkinson
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this library; if not, write to the Free Software
+# Foundation, 51 Franklin Street, Suite 500 Boston, MA 02110-1335 USA
+'''
+The filetransfer module provides a high level API accepting and
+sending file transfers to/from either memory of real files.
+
+.. note ::
+    This module provides options that activities should not use as they
+    will trigger functions in the shell.  Activities should use the
+    file transfer APIs in the :class:`sugar3.presence.wrapper.CollabWrapper`.
+'''
+
+import os
+import socket
+
+from gi.repository import GObject
+from gi.repository import Gio
+from gi.repository import GLib
+import dbus
+
+from telepathy.interfaces import \
+    CHANNEL_TYPE_FILE_TRANSFER, \
+    CHANNEL
+from telepathy.constants import \
+    CONNECTION_HANDLE_TYPE_CONTACT, \
+    SOCKET_ADDRESS_TYPE_UNIX, \
+    SOCKET_ACCESS_CONTROL_LOCALHOST
+from telepathy.client import Channel
+
+import logging
+
+FT_STATE_NONE = 0
+FT_STATE_PENDING = 1
+FT_STATE_ACCEPTED = 2
+FT_STATE_OPEN = 3
+FT_STATE_COMPLETED = 4
+FT_STATE_CANCELLED = 5
+
+FT_REASON_NONE = 0
+FT_REASON_REQUESTED = 1
+FT_REASON_LOCAL_STOPPED = 2
+FT_REASON_REMOTE_STOPPED = 3
+FT_REASON_LOCAL_ERROR = 4
+FT_REASON_LOCAL_ERROR = 5
+FT_REASON_REMOTE_ERROR = 6
+
+
+class BaseFileTransfer(GObject.GObject):
+    '''
+    The base file transfer should not be used directly.  It is used as a
+    base class for the incoming and outgoing file transfers.
+    '''
+
+    filename = None
+    '(str): metadata provided by the buddy'
+
+    description = None
+    '(str): metadata provided by the buddy'
+
+    file_size = None
+    '(int): size of the file being sent/received, in bytes'
+
+    mime_type = None
+    '(str): metadata provided by the buddy'
+
+    buddy = None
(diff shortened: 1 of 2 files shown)


samdroid-apps:
Preview of documentation on this branch:  http://people.sugarlabs.org/sam/docs-collab-wrapper-try2/

Collab wrapper class:  http://people.sugarlabs.org/sam/docs-collab-wrapper-try2/sugar3.presence.wrapper.html

File transfer abstractions:  http://people.sugarlabs.org/sam/docs-collab-wrapper-try2/sugar3.presence.filetransfer.html

Another item for the TODO is to port jarabe to use the file transfer api in this patch rather than in jarbae.model.filetransfer.  However, that is simple compared to this patch, and could eaisly be done later as both apis are very simmilar.

samdroid-apps:
Great, ran through some more tests and resuming a saved activity works fine.  I tested the following test case:

1.  Alice opens up a previously shared saved activity
2.  Alice invites Bob
3.  Bob joins and adds new content
4.  Alice sees the new content

It doesn't work if 2 people resume the shared activity from their journals.  Is that meant to work?  I think it is probably an issue in the code that decides who will lead the activity.

walterbender:
We should check, but I always thought that the sharer (leader) was not
something that could be handed off to other participants. There are
countless corner cases that never really got sorted out definitively.

On Sun, Dec 6, 2015 at 3:22 PM, Sam <notifications@github.com> wrote:

> Great, ran through some more tests and resuming a saved activity works
> fine. I tested the following test case:
>
>    1. Alice opens up a previously shared saved activity
>    2. Alice invites Bob
>    3. Bob joins and adds new content
>    4. Alice sees the new content
>
> It doesn't work if 2 people resume the shared activity from their
> journals. Is that meant to work? I think it is probably an issue in the
> code that decides who will lead the activity.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/272#issuecomment-162344616>
> .
>



-- 
Walter Bender
Sugar Labs
http://www.sugarlabs.org

samdroid-apps:
oh, thats not good.

I only use the leader in this wrapper to deal with initial state sync.  However, that could probably replaced with all initized buddies offering to send the state to the noob, and the noob only accepting 1 file transfer.

But that still doesn't help with the resuming shared activity thing I don't think.  Do we need to sync state when people resume a shared activity?  I'm not sure how to implement it if we do because how do we know if the buddy is the first to come to the shared activity (and should be marked initized) or is joining a shared activity by resuming it?????  I'm very confused right now/

walterbender:
Don't be confused. Some of the corner cases -- which are largely
activity-specific -- were not ever resolved.

On Sun, Dec 6, 2015 at 3:45 PM, Sam <notifications@github.com> wrote:

> oh, thats not good.
>
> I only use the leader in this wrapper to deal with initial state sync.
> However, that could probably replaced with all initized buddies offering to
> send the state to the noob, and the noob only accepting 1 file transfer.
>
> But that still doesn't help with the resuming shared activity thing I
> don't think. Do we need to sync state when people resume a shared activity?
> I'm not sure how to implement it if we do because how do we know if the
> buddy is the first to come to the shared activity (and should be marked
> initized) or is joining a shared activity by resuming it????? I'm very
> confused right now/
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/272#issuecomment-162346132>
> .
>



-- 
Walter Bender
Sugar Labs
http://www.sugarlabs.org

quozl:
Typo in commit description first line.

samdroid-apps:
@quozl sorry, I don't see it.

Any other comments on this patch?

samdroid-apps:
For reference, here is the patch for browse:  https://github.com/sugarlabs/browse-activity/pull/17

quozl:
@samdroid-apps wrote:
> @quozl sorry, I don't see it.

It isn't how we spell collaboration.  Also in remainder of commit.  In sed terms, s/collaberation/collaboration/g

quozl:
I'm yet to see a feature page, and once I do there'll be a ton of questions. But looking at the patch for browse makes me think "CollabWrapper" as a name won't be so easily understood.  Perhaps give it a name that is shorter, and that doesn't expose implementation detail.  Go look for synonyms of collaboration.

samdroid-apps:
Ok, you can see a feature page:  https://wiki.sugarlabs.org/go/Features/Fixing_Collab_(Tubes)

If you have any preferences for another name feel free to suggest.

src/sugar3/presence/filetransfer.py
```diff
@@ -0,0 +1,358 @@
...
+    def __init__(self, buddy, conn, path, filename, description, mime):
+        BaseOutgoingTransfer.__init__(
+            self, buddy, conn, filename, description, mime)
+
+        self._path = path
+        file_size = os.stat(path).st_size
+        self._create_channel(file_size)
+
+    def _get_input_stream(self):
+        logging.debug('opening %s for reading', self._file_name)
```
samdroid-apps: Oops, this is an issue.  I'll fix this up an prepare a patch for image viewer

samdroid-apps:
Development moved to https://github.com/samdroid-apps/collabwrapper

Maybe I'll try to merge this next cycle

nullr0ute:
Can we call it sugar-collabwrapper if it needs to be packaged separately, I'd sooner it be part of toolkit or the core sugar though. Does it mean it's not being fixed the 107 cycle (again)?

samdroid-apps:
No more packages will be requires, as the code in included in activities git repos (sometimes as submodules, other times just copied and pasted).

The new api for activities will not be included in sugar3 during the 10{7,8} cycle.  However some activities have began using it by including a copy of the  https://github.com/samdroid-apps/collabwrapper repository, which will allow for faster (no timeline) development between now and the next cycle.

PR #275, opened 2015-12-14
demote warning to debug for sugar_accel_group

quozl:
This warning is frequently seen in logs but is not that important to warrant being a warning.


tchx84:
Reviewed and merged.

PR #276, opened 2015-12-15
Fix remaining icon_size references

quozl:
icon_size was changed to pixel_size, yet references remained, causing
extra log data.


godiard:
Sorry, I merged too fast. 
You need change the constant defining the size too, see https://developer.sugarlabs.org/release-notes/102.md.html#icon%20size

PR #279, opened 2015-12-21
Fix persistent tree view palettes

quozl:
A rework of the TreeViewInvoker used by the journal and activity list.

Fixes a persistent journal entry palette that reappears in journal or home view.  See [#4918](https://bugs.sugarlabs.org/ticket/4918).

Also fixes an over-sensitive journal icon; with the journal mostly empty, rapidly move the mouse from the bottom of the screen into the lowest icon and then out again.  Before this patch, the palette was shown.  After this patch, the mouse must rest in the icon.

Changes made:

- expose _ensure_palette_exists in Invoker and rename to force_palette, so it can be used in TreeViewInvoker,

- remove reference to MouseSpeedDetector, it is no longer required,

- remove enter and leave events; they are for the TreeView as a whole, and are not required,

- properly detect that None is returned by get_path_at_pos, when the coordinates are outside the TreeView cells.

Not fixed:

- moving the mouse out of the journal icon invoker does not popdown the palette, but it does when moving the mouse out of the activity icon in the activity list; so this is a problem beyond the invoker,

Tested on Fedora 18 and Ubuntu 15.10.

Test cases:

- check a click will activate item,

- check a right-click will show palette,

- check a mouse hover will show palette,

- check a brief mouse passing will not show palette,

- check if mouse leaving palette will popdown.

### DIFF
--- a/src/sugar3/graphics/palettewindow.py
+++ b/src/sugar3/graphics/palettewindow.py
@@ -1411,14 +1411,11 @@ class TreeViewInvoker(Invoker):
 
         self._tree_view = None
         self._motion_hid = None
-        self._leave_hid = None
         self._release_hid = None
         self._long_pressed_hid = None
         self._position_hint = self.AT_CURSOR
 
         self._long_pressed_controller = SugarGestures.LongPressController()
-
-        self._mouse_detector = MouseSpeedDetector(200, 5)
 
         self._tree_view = None
         self._path = None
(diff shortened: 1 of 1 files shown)


src/sugar3/graphics/palettewindow.py
```diff
@@ -995,14 +995,14 @@ def notify_popdown(self):
         self._cursor_x = -1
         self._cursor_y = -1
 
-    def _ensure_palette_exists(self):
+    def force_palette(self):
```
samdroid-apps: Why did you rename it force_palette?  That seems unnecessary, as ensure_palette_exsists is very descriptive.

Either way, if you're renaming a method and making it public, you should consider documenting it :)

samdroid-apps:
Wow, this is great!  It also fixes the following test case:

1.  Mouse over a jobject palette in the icon view
2.  Mouse out
Expected result:  Palette goes away after a short timeout  
Actual result (without patch):  Palette stays forever

samdroid-apps:
Oh, just saw that test case was noted in your not fixed section.  Still a merge for me, although feel free to force push documentation for the function that you made public.

quozl:
Thanks.  Force pushed:
- avoid rename of method,
- avoid making a method public, now private to file,
- noted _ensure_palette_exists is overridden by TreeViewInvoker,
- tested that __motion_notify_event_cb does call the overridden method through notify_mouse_enter.

PR #282, opened 2015-12-24
Only give activities telepathy tubes if marked as supported

samdroid-apps:
Requires https://github.com/sugarlabs/sugar/pull/644

Tubes API has been deprecated since 0.17.25 [1], which was probably
around 2009.  This has since been removed from telepathy, breaking
collaberation.

To maintain support for older activities, if a gsetting is set,
tubes will still be given.  This is useful for OLPC OS, which is
running an older telepathy version that still supports telepathy
tubes.  This therefore requires a patch for the gsettings schema.

When enabled, activities are still given a text channel.  They can
also make their own channels by using the telepathy connection and
calling CreateChannel on the requests interface.

### DIFF
--- a/src/sugar3/presence/activity.py
+++ b/src/sugar3/presence/activity.py
@@ -27,6 +27,7 @@ from functools import partial
 import dbus
 from dbus import PROPERTIES_IFACE
 from gi.repository import GObject
+from gi.repository import Gio
 from telepathy.client import Channel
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Hum, this assumes that tubes are supported/not for both telepathy salut and gabble.  However, it appears that tubes are currently only removed in gabble.  This patch will need to be redone with that in mind.

PR #291, opened 2015-12-30
Write docs for sugar3.graphics.scrollingdetector

ndefilippis:


src/sugar3/graphics/scrollingdetector.py
```diff
@@ -15,16 +15,35 @@
...
-    """
+    '''
+    The scrolling detector sends signals when a scrolled window is scrolled and 
+    when a scrolled window stops scrolling
+    
+    The `scroll-start` signal is emitted when scrolling begins and
+    The `scroll-end` signal is emitted when scrolling ends
+    
+    Args:
+        scrolled_window (gtk.ScrolledWindow): A GTK scrolled window object for which
```
samdroid-apps: Gtk (lowercase g usually refers to pygtk - which is the older version)

src/sugar3/graphics/scrollingdetector.py
```diff
@@ -15,16 +15,35 @@
...
-    ScollingDetector emit signals when a ScrolledWindow starts and
-    finish scrolling. Other widets can use that information to
-    avoid do performance expensive operations.
-    """
+    '''
+    The scrolling detector sends signals when a scrolled window is scrolled and 
+    when a scrolled window stops scrolling
+    
+    The `scroll-start` signal is emitted when scrolling begins and
+    The `scroll-end` signal is emitted when scrolling ends
```
samdroid-apps: Please document that neither signals have any arguments

src/sugar3/graphics/scrollingdetector.py
```diff
@@ -39,10 +58,23 @@ def __init__(self, scrolled_window, timeout=100):
         GObject.GObject.__init__(self)
 
     def connect_scrolled_window(self):
+        '''
+        Connects scrolling detector to a scrolled window.
+        Detects scrolling when the vertical scrollbar
+        adjustment value is changed
```
samdroid-apps: When should I use this function?  Please document that

ndefilippis:
I've applied these changes

quozl:
I don't think there should be documentation for internal methods of the class, at least not in a form that suggests the methods can be used by the caller.

ndefilippis:
Won't sphinx automatically exclude private methods?

ndefilippis:
Would it make sense to put the description of the private methods in a regular comment, rather than a doc string?

quozl:
I'm puzzled as to how conenct_scrolled_window could be used.  Since it depends on self._scrolled_window, which is set by __init__, and since it is called by __init__, it doesn't look like it has any purpose.  @samdroid-apps, what do you think?

quozl:
@godiard, you added this in 5626a6c182ba4a8fca3c750e38ea144264a813f2, what do you think?

godiard:
@quozl, true. Probably the method should be private or the code included in __init__().
Anyway, if the change is done, should be in a patch different to the documentation patch

src/sugar3/graphics/scrollingdetector.py
```diff
@@ -15,16 +15,37 @@
 # Free Software Foundation, Inc., 59 Temple Place - Suite 330,
 # Boston, MA 02111-1307, USA.
 
+'''
+ScrollingDetector emits signals when a ScrolledWindow starts and
+finish scrolling. Other widgets can use that information to
+avoid doing performance-expensive operations.
+
+Example:
+.. literalinclude:: ../examples/scrollingdetector.py
```
samdroid-apps: Incorrect syntax, I think that you need a newline aftter the example.

ndefilippis:
Should work now

PR #269, opened 2015-11-14
Write documentation for sugar3.graphics.combobox

samdroid-apps:

PR #294, opened 2015-12-31
Write docs for sugar3.graphics.notebook

ndefilippis:


ndefilippis:
Changes made

src/sugar3/graphics/notebook.py
```diff
@@ -15,21 +15,23 @@
...
+'''
+Notebook class
 
-This class create a Gtk.Notebook() widget supporting
-a close button in every tab when the 'can-close-tabs' gproperty
-is enabled (True)
-
-STABLE.
-"""
+This class creates a Gtk.Notebook() widget supporting
```
samdroid-apps: :class:`Gtk.Notebook`

src/sugar3/graphics/notebook.py
```diff
@@ -15,21 +15,23 @@
...
+'''
 
 from gi.repository import Gtk
 from gi.repository import GObject
 
 
 class Notebook(Gtk.Notebook):
-
+    '''
+    Notebook class that creates a Gtk.Notebook. It is possible to set
```
samdroid-apps: Same here

samdroid-apps: How do you set the can close tabs prop from the contructor?  Maybe give an example like (untested):

    Notebook(can_close_tabs=false)

I think that is what the kwargs are doing

src/sugar3/graphics/notebook.py
```diff
@@ -114,21 +102,16 @@ def _create_custom_tab(self, text, child):
...
 
-        Returns
-        -------
-        Boolean
-          Returns TRUE if the page is successfully added to the notebook.
+        Args:
+            text_label (string): label of page to be added
 
-        """
+            widget (.. :class:`Gtk.Widget`): widget to be used as contents of current page
```
samdroid-apps: I think the syntax is wrong, it believe that you just need (Gtk.Widget)

quozl:
Thanks.

I didn't know from src/sugar3/graphics/notebook.py (after your patch) that a keyword argument was needed to set the "can-close-tabs" at instantiation.

```
nb = Notebook(can_close_tabs=True)
```

I needed to refer to example/notebook.py to find out, so perhaps that could be clearer in the API documentation.

Perhaps relevant parts of examples/notebook.py should be in the API documentation.

I see that apart from examples/notebook.py there are no other uses of this API, so that constrains my review.

I agree with merge now; it is an improvement on what we have.  It can also be better.

samdroid-apps:
Please include the example in the introduction section of the documentation

ndefilippis:
Can that be done using .. literalinclude?

samdroid-apps:
Yes, eg. https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/269/files#diff-3cbf9d7bdb3a17e2a1f947d871ea283cR23

ndefilippis:
Changes made

quozl:
Reviewed.  No further comment.

src/sugar3/graphics/notebook.py
```diff
@@ -15,21 +15,25 @@
...
-This class create a Gtk.Notebook() widget supporting
-a close button in every tab when the 'can-close-tabs' gproperty
-is enabled (True)
+This class creates a :class:`Gtk.Notebook` widget supporting
+a close button in every tab when the `can-close-tabs` gproperty
+is enabled (True).
 
-STABLE.
-"""
+.. literalinclude ../examples/notebook.py
```
samdroid-apps: For some reason this doesn't show up when sphinx renders the docs

[https://gyazo.com/165e853a3ed7dd8572aa30385ba0f240](https://gyazo.com/165e853a3ed7dd8572aa30385ba0f240)

ndefilippis:
Should work now

PR #295, opened 2016-01-01
Write docs for sugar3.graphics.radiotoolbutton

ndefilippis:


src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -16,9 +16,15 @@
...
-STABLE.
-"""
+'''
+Provides a RadioToolButton class, similar to a "push" button.
+When the button is clicked, it depresses and is shaded darker.
+It is also possible to set a tooltip to be dispalyed when the user
+scrolls over it with their cursor.
+
+Example:
+    .. literalinclude ../examples/radiopalette.py
```
samdroid-apps: That example is very old and doesn't even run

```
  File "radiopalette.py", line 13, in <module>
    box.pack_start(toolbar, False)
TypeError: Gtk.Box.pack_start() takes exactly 5 arguments (3 given)
```

Please make a more consise example of just the radiotoolbuttons, maybe looking at how they are used in the frame or homeview code

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -29,7 +35,13 @@
...
 
 class RadioToolButton(Gtk.RadioToolButton):
-    """An implementation of a "push" button."""
+    '''
+    An implementation of a "push" button. Used with a :class:`RadioPalette` to create a group of 
+    RadioToolButtons.
+    
+    Args:
+        icon_name (string): name of icon to be used
+    '''
```
samdroid-apps: Hum don't think you need a radio palette.  Eg from jarabe/frame/zoomtoolbar.py it uses the "group" kwarg.  That is just another RadioToolButton that's in the same group

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -71,28 +89,49 @@ def set_tooltip(self, tooltip):
...
+        Returns the tooltip
+        '''
         return self._tooltip
 
     tooltip = GObject.property(type=str, setter=set_tooltip,
                                getter=get_tooltip)
 
     def set_accelerator(self, accelerator):
+        '''
+        Sets accelerator
```
samdroid-apps: Please explain

src/sugar3/graphics/radiotoolbutton.py
samdroid-apps: Maybe you should note the names of the gobject props (which you can pass as args to the constructor)

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -102,12 +141,21 @@ def get_icon_name(self):
                                  getter=get_icon_name)
 
     def set_xo_color(self, xo_color):
+        '''
+        Sets xocolor
+        
+        Args:
+            xo_color (sugar3.graphics.xocolor): xocolor to be set
```
samdroid-apps: sugar3.graphics.xocolor is the module, you want to add XoColor to refer to the class

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -102,12 +141,21 @@ def get_icon_name(self):
                                  getter=get_icon_name)
 
     def set_xo_color(self, xo_color):
+        '''
+        Sets xocolor
```
samdroid-apps: Please elaborate.  Xocolor of what?

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -117,18 +165,37 @@ def create_palette(self):
         return None
 
     def get_palette(self):
+        '''
+        Returns this radiotoolbutton's palette. A palette groups
```
samdroid-apps: No, a palette is like the right click menu.

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -117,18 +165,37 @@ def create_palette(self):
...
+            palette (sugar3.graphics.RadioPalette)
+        '''
         self._palette_invoker.palette = palette
 
     palette = GObject.property(
         type=object, setter=set_palette, getter=get_palette)
 
     def get_palette_invoker(self):
+        '''
+        Returns palette invoker
```
samdroid-apps: Can you find an example of this?  Also, can you explain this a little more?

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -152,9 +222,19 @@ def do_draw(self, cr):
         return False
 
     def get_hide_tooltip_on_click(self):
+        '''
+        Returns True if the tooltip is hidden when a user clicks on it, otherwise returns false
```
samdroid-apps: it?  which it?  the radio tool button?

examples/radiotoolbutton.py
```diff
@@ -0,0 +1,40 @@
+from gi.repository import Gtk
+
+from sugar3.graphics.radiotoolbutton import RadioToolButton
+from sugar3.graphics.radiopalette import RadioPalette, RadioMenuButton
+from sugar3.graphics.xocolor import XoColor
+import common
+
+test = common.Test()
```
samdroid-apps: I'm not a big an of common.test, because it makes more sense to just use a Gtk.Window so people can copy the code.

So just something like

    window = Gtk.Window()
    window.show()
    window.connect('destroy', Gtk.main_quit)

then add the box to window

and instead of common.main, just run `Gtk.main()`

examples/radiotoolbutton.py
```diff
@@ -0,0 +1,40 @@
...
+box = Gtk.HBox()
+test.pack_start(box, True, True, 0)
+box.show()
+
+def echo(button, label):
+    if not button.props.active:
+        return
+    print label
+
+palette = RadioPalette()
```
samdroid-apps: Please define palette down where it is used

examples/radiotoolbutton.py
```diff
@@ -0,0 +1,40 @@
...
+
+button1 = RadioToolButton(icon_name='document-save', accelerator="<ctrl>S", xo_color=XoColor("#FF0000,#00FF00"), tooltip="save")
```
samdroid-apps: This line is a little long.

Try running pep8 (or flake8) on your code)

samdroid-apps: Please don't hardcode XoColor values.  Use one from graphics.style or just use the default

examples/radiotoolbutton.py
```diff
@@ -0,0 +1,40 @@
...
+button1.show()
+
+button2 = RadioToolButton(icon_name='document-open', accelerator="<ctrl>O", xo_color=XoColor("#0000FF,#FF0000"), tooltip="open")
+button2.connect('clicked', lambda button: echo(button, 'document-open'))
+box.pack_start(button2, False, False, 1)
+button2.show()
+
+button3 = RadioToolButton(icon_name='document-send', accelerator="<ctrl>F", xo_color=XoColor("#0000FF,#00FF00"))
+button3.connect('clicked', lambda button: echo(button, 'document-send'))
+palette.append(button3, 'menu.document-send')
```
samdroid-apps: What is your code doing here?  I'm confused!  Please use comments to explain things that aren't the most obvious

examples/radiotoolbutton.py
```diff
@@ -0,0 +1,40 @@
...
+button = RadioMenuButton(palette=palette)
+box.pack_start(button, False, False, 1)
+button.show()
```
samdroid-apps: Good example, but you didn't actually use the RadioToolButton features (the fact that you can have a bunch of them and select only 1 at the time).  Please look into that, although this could be a good ToolButton example

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -29,7 +35,27 @@
...
+	be set are:
+		accelerator (string)
+		tooltip (string)
+		xocolor (:class:`XoColor`)
+		hide_tooltip_on_click (bool)
+    
+    Args:
+        icon_name (string): name of icon to be used
+		
+		kwargs can be used to set the following properties
```
samdroid-apps: You can document kwargs by just using the Keyword Args: header and then using the same syntax as normal args

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -29,7 +35,27 @@
 
 
 class RadioToolButton(Gtk.RadioToolButton):
-    """An implementation of a "push" button."""
+    '''
+    An implementation of a "push" button. Properies that can
```
samdroid-apps: GObject properties, not normal properties I think

samdroid-apps: Also, I don't think that the description really explains that these are radio buttons (not just ToolButtons)

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -117,18 +179,36 @@ def create_palette(self):
...
+        Returns this button's palette. 
+        '''
         return self._palette_invoker.palette
 
     def set_palette(self, palette):
+        '''
+        Sets the palette
+        
+        Args:
+            palette (:class:`RadioPalette`)
```
samdroid-apps: just any palette?

ndefilippis:
Do you think that the palette_invoker methods are implementation methods and not meant to be called?

ndefilippis:
Also, graphics.style does'nt have any xocolor constants, just normal colors. How would I use the default?

examples/radiotoolbutton.py
```diff
@@ -0,0 +1,47 @@
...
+    if not button.props.active:
+        return
+    print label
+
+
+palette = RadioPalette()
+# Adding 3 RadioToolButtons to a palette
+
+button1 = RadioToolButton(icon_name='document-save', accelerator="<ctrl>S", /
+                          xo_color=XoColor("#FF0000,#00FF00"))
```
samdroid-apps: Maybe use "white" or "insensitive" as the XoColor string?

examples/radiotoolbutton.py
```diff
@@ -0,0 +1,47 @@
...
+button1.connect('clicked', lambda button: echo(button, 'document-save'))
```
samdroid-apps: Isn't there a special signal that is used for the radiotoolbuttons that detects when they are toggled or sonething?

Have a look at https://github.com/samdroid-apps/sugar/blob/help-nav-buttons/src/jarabe/view/viewhelp.py#L307

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -29,7 +39,29 @@
...
+        icon_name (string): name of icon to be used
+
+    Keyword Args:
+
+        accelerator (string): keyboard shortcut to be used to
+        activate this button
+
+        tooltip (string): tooltip to be displayed when user hovers over button
+
+        xo_color (:class:`XoColor`): XoColor of button
```
samdroid-apps: Just a styntax thing, it only needs to be `(sugar3.graphics.xocolor.XoColor)`, no special `:class:`

src/sugar3/graphics/radiotoolbutton.py
```diff
@@ -16,9 +16,19 @@
...
+A group of RadioToolButtons can be set, so that only one can be
+selected at a time. When a button is clicked, it depresses and
+is shaded darker.
+
+It is also possible to set a tooltip to be dispalyed when the
+user scrolls over it with their cursor as well as an accelerator
+keyboard shortcut.
+
+Example:
+    .. literalinclude ../examples/radiotoolbutton.py
```
samdroid-apps: Oh, this didn't build correctly.  Please test that in the future.

It should be:

```
Example:                                                                        
                                                                                
.. literalinclude:: ../examples/radiotoolbutton.py   
```

ndefilippis:
Changes made

samdroid-apps:
I think this is already merged.  It is already in the codebase at least.  Closing.

PR #296, opened 2016-01-04
Handle directories in bundle builder install, fixes git submodules

samdroid-apps:
Currently, having a submodule in an activity crashes the install
script as git returns submodules as directories in git ls-files.
This commit means that submodules are reccursively coppied, as
they would be with "cp -r".

To reproduce the issue, you could do something like the following:

    cd pippy
    git clone https://github.com/samdroid-apps/collabwrapper
    git add collabwrapper
    python setup.py install  # or just osbuild run

### DIFF
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -283,7 +283,15 @@ class Installer(Packager):
             if not os.path.exists(path):
                 os.makedirs(path)
 
-            shutil.copy(source, dest)
+            if os.path.isfile(source):
+                shutil.copy(source, dest)
+            else:
+                if os.path.isdir(dest):
+                    # shutil can't copytree is dest exists
+                    shutil.rmtree(dest)
+                # Git may give us directories to copy, namely with git
+                # submodules they are only included as their root path
+                shutil.copytree(source, dest)
 
         if install_mime:
             self.config.bundle.install_mime_type(self.config.source_dir)


quozl:
Reviewed.

- commit description first line might be "Install git submodules in bundlebuilder",

- commit description body typo; *coppied* should be *copied*,

- what about symbolic links; perhaps the first ```os.path.isfile``` should be ```not os.path.isdir```,

- first comment seems excessive,

- second comment seems misplaced; might be moved to before the code block, or removed instead,

- if you think the recursive copy of this code block should be thoroughly documented, perhaps move it to a new method?

samdroid-apps:
Thanks for the review.

I will send a new patch with your comments addressed.  I will also expand the scope to include an unknown change to the "dist_xo" command, as that does not work as expected either

samdroid-apps:
Also I believe that it may be beneficial to resolve symbolic links instead of assuming they are files.  Would that be ok?

quozl:
Hmm, interesting question.  Our bundle format .xo is based on zip, and zip does not represent symbolic links properly.  I've tested;
- links are followed to their files and the files are placed in the container,
- links that can't be followed are omitted from the container.

So it would seem following them might be best.

godiard:
I see a potential problem with including submodules. Until now, activities code was always in a single repository, then was easy to young developers clone and start coding. If we allow the use of submodules, or simlinks, will be more complex, and we need document in the activities what is the procedure to get the sources. And this do not solve the issue of change a submodule that need changes in the activity and all stay in a inconsistent state.

PR #297, opened 2016-01-04
collaboration: handle tubes channel failure gracefully

tchx84:
CHANNEL_TYPE_TUBES was deprecated since telepathy-gabble 0.17.25 and
is no longer present on modern distributions. This breaks collaboration
for _all_ activities, regardless if they use tubes or not.

Therefore, check if the channel creation failed due to support and allow
channel creation process to continue.

This patch fixes collaboration for the activities that do not use tubes
channel in modern distributions, but also allow to use tubes when running
on older distributions.

### DIFF
--- a/src/sugar3/presence/activity.py
+++ b/src/sugar3/presence/activity.py
@@ -554,6 +554,7 @@ class _JoinCommand(_BaseCommand):
         self._finished = False
         self.room_handle = room_handle
         self._global_self_handle = None
+        self._tubes_supported = True
 
     def run(self):
         if self._finished:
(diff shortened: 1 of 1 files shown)


tchx84:
This replaces https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/282 and https://github.com/sugarlabs/sugar/pull/644.

samdroid-apps:
I have merged this patch.  I have done limited testing (fc24 with salut and gabble) and the patch successful runs chat (no tubes) on gabble (tubes unsupported), and runs turtle (tubes dependant) on salut (tubes supported).

Hopefully distro maintainers can package up the next unstable release so that this patch can be tested.

PR #298, opened 2016-01-05
Handle git submodules in bundlebuilder

samdroid-apps:
Extends the fix of #296

Git submodules are only listed by their root directory in the
output of "git ls-files".  Therefore, we must handle this case
so that the files from the module are included in the install
and dist_xo commands.

To reproduce the issue, you could do something like the following:

    cd pippy
    git clone https://github.com/samdroid-apps/collabwrapper
    git add collabwrapper
    python setup.py install  # or just osbuild run

### DIFF
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -168,12 +168,15 @@ class Packager(object):
         if not os.path.exists(self.config.dist_dir):
             os.mkdir(self.config.dist_dir)
 
-    def get_files_in_git(self):
+    def get_files_in_git(self, root=None):
+        if root is None:
+            root = self.config.source_dir
+
         git_ls = None
         try:
             git_ls = subprocess.Popen(['git', 'ls-files'],
                                       stdout=subprocess.PIPE,
-                                      cwd=self.config.source_dir)
+                                      cwd=root)
         except OSError:
             logging.warn('Packager: git is not installed, '
                          'fall back to filtered list')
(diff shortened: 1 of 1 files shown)


godiard:
Please see my comment in #296

samdroid-apps:
@godiard said on the other pull request:

> I see a potential problem with including submodules. Until now, activities code was always in a single repository, then was easy to young developers clone and start coding. If we allow the use of submodules, or simlinks, will be more complex, and we need document in the activities what is the procedure to get the sources. And this do not solve the issue of change a submodule that need changes in the activity and all stay in a inconsistent state.

Symbolic links are a bad idea as Quozl said on the other patch, apologies for my naivity of suggesting them.  They are not handled in this patch.

Git submodules are more complex than straight git as you say.  However it solves an important issue for collab wrapper which is "how do we not end up with groupthink 2.0".  (groupthink is a collab framework used in pippy and some others).  Groupthink seems pretty bad as it has been copied and pasted between activities, whereas it should be a project of it's own.  Submodules fix this issue for collab wrapper by allowing activities to not have to copy and paste lots of code to use it.  It also allows for collab wrapper to change, but activities can pin the version that they use.  Ultimately, collabwrapper should be merged, however this patch helps fill the gap.

This is especially important now that collabwrapper has expanded to 2 files with the collaberative text buffer/editor.  This in particular is gonna be in implementation flux as it matures and handles more edge cases.

I believe that cutting down code copy and paste is more important than avoiding submodules.

godiard:
@samdroid-apps, both approach have good and bad points. I suggest you propose the change and discuses the implications in mailing list. This is not a simple fix, change important assumptions in how the activities are developed.

quozl:
Discussing and merging this is becoming more urgent, as things are breaking.  Pippy-68 released via ASLO has no collabwrapper in the bundle despite https://github.com/walterbender/Pippy/commit/8f184ab637953007b3cab013d1e9ff2247cea6b7 presumably because it was built on a system without #298 applied.  Therefore Pippy-68 fails to start.

cc @walterbender.

samdroid-apps:
This could be the time to discuss and act now, as we are in a feature window!

samdroid-apps:
Given the lack of objections, this will be part of 0.109.0.1

PR #301, opened 2016-01-12
Update toggletoolbutton.py

Hrishi1999:


src/sugar3/graphics/toggletoolbutton.py
```diff
@@ -28,13 +27,18 @@
...
+Toogle tool button switches you between two tasks.
+For example, a button can contain two functions like
+On or Off.
+'''
+
 
 def _add_accelerator(tool_button):
     if not tool_button.props.accelerator or not tool_button.get_toplevel() or \
             not tool_button.get_child():
         return
```
goutamnair7: Is there any reason why you removed this TODO ?

goutamnair7:
It looks fine to me. @quozl please review.

Hrishi1999:
@quozl Can you review, please?

src/sugar3/graphics/toggletoolbutton.py
Hrishi1999: It must have been deleted accidentally while pep8'ing.

goutamnair7: It would be a good idea to keep it, in the case that it is not resolved yet.

quozl:
Reviewed.

The proposed documentation explains the class so very badly that I think @Hrishi1999 has never written a Gtk3 program in their life, or english language skills are very weak.

Remember that writing documentation for a Python class can only be done effectively by someone who can not only write a program that uses the class, but change the class to meet new requirements.  It is not a task for beginners.

The same skills are needed by reviewers.

It would be faster for me to write this documentation myself than to explain how to write it, but I'll struggle on regardless. :grin:

For the class ToggleToolButton;
 - there are no tasks, yet you introduce the term,
 - there are no functions on or off, yet you introduce the term,

Other review notes;
 - you have copied Gtk.ToggleToolButton documentation, making it look like the class ToggleToolButton is named Gtk.ToggleToolButton, and further you used the [C language documentation](https://developer.gnome.org/gtk3/stable/GtkToggleToolButton.html) not the [Python language documentation](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/ToggleToolButton.html),
 - you have listed keyword arguments for the class, yet the only keyword argument in the ```__init__``` declaration is ```icon_name```; if you don't know what keyword arguments are, please read the Python language documentation,
 - your documentation of the ```set_icon_name``` method is a copy of Gtk2 documentation for the ```set_icon_name``` method of the [gtk.ToolButton class](http://www.pygtk.org/pygtk2reference/class-gtktoolbutton.html#method-gtktoolbutton--set-icon-name), instead of the [Gtk3 documentation](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/ToolButton.html#Gtk.ToolButton.set_icon_name),
 - you still have an indentation error in ```get_accelerator```, where you changed the executable code, which would probably break our build.

For your interest, the Sugar ToggleToolButton class is a Gtk.ToggleToolButton with an icon, a tooltip palette, and an accelerator.

Hrishi1999:
Hello, i am doing this task because of course i know python.
I know English. The on off thing was just an example. I am not a beginner. 
Anyway, i will fix it

quozl:
Good, thanks.

src/sugar3/graphics/toggletoolbutton.py
```diff
@@ -28,6 +27,12 @@
 from sugar3.graphics.icon import Icon
 from sugar3.graphics.palette import Palette, ToolInvoker
 
+'''
+Toogle tool button switches you between two tasks.
+For example, a button can contain two functions like
+On or Off.
```
quozl: I still don't think "tasks" and "on or off" is a good example.  If you feel you must use an example to explain this Sugar ToggleToolButton because it is different to Gtk.ToggleToolButton, then use the Journal favourites selection button as a guide.

quozl:
Reviewed.  Hope others will review now.

Hrishi1999:
Thanks.

goutamnair7:
@Hrishi1999 Please squash your commits into a single one

PR #306, opened 2016-01-20
Remove Rainbow, avoid race in creating directories

quozl:
Sugar depended on Rainbow for clearing the activity instance/ and tmp/ directories.  But Rainbow is no longer used downstream.

 - remove support for Rainbow,

 - avoid race when creating directories; don't check they exist before we create.

### DIFF
--- a/src/sugar3/activity/activityfactory.py
+++ b/src/sugar3/activity/activityfactory.py
@@ -38,6 +38,7 @@ import os
 import tempfile
 import subprocess
 import pwd
+import shutil
 
 _SHELL_SERVICE = 'org.laptop.Shell'
 _SHELL_PATH = '/org/laptop/Shell'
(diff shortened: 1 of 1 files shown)


samdroid-apps:
This patch is a NAK from my perspective.

As @godiard said on the mailing list, what if the user has 2 activity instances open?  When you close the first instance, it will clear it's instance dir and leave it in a bad state.  Please clear it at a different time to address that issue.

Otherwise it looks like a good patch.

quozl:
Any reason why not say this on mailing list where the discussion is?

quozl:
Ping?  Substantial change to patch, please review.

samdroid-apps:
From Gonzalo's reply to your initial mailing list post:

> A few issues:
> ...
> * Temporary directories are shared between instances, if we remove temporary directories 
at activity start or stop, we need check if there are other instances of the activity running.

Please address this.  Currently, your patch description says it will delete files from under the activity's eyes if a 2nd instance is launched.

Also, where is the code that actually removes the directories?

quozl:
> Currently, your patch description says it will delete files

No, it doesn't.  Look at the commit carefully?

samdroid-apps:
Ah, I see.  I just edited the initial pull request comment to reflect the new commit message.

This looks fine to me.  I'll just test 'n' merge.

PR #309, opened 2016-02-24
progresstoolbutton.py added in src/graphics/

iamutkarshtiwari:
@samdroid-apps This PR integrated this requested feature -> https://bugs.sugarlabs.org/ticket/4721
Cross reference to other required PR - https://github.com/sugarlabs/browse-activity/pull/24
desktop-animation

### DIFF
--- a/src/sugar3/graphics/Makefile.am
+++ b/src/sugar3/graphics/Makefile.am
@@ -16,6 +16,7 @@ sugar_PYTHON =              \
 	palettemenu.py          \
 	palettewindow.py        \
 	panel.py                \
+	progresstoolbutton.py   \
 	radiopalette.py         \
 	radiotoolbutton.py      \
 	scrollingdetector.py    \
--- a/src/sugar3/graphics/progresstoolbutton.py
+++ b/src/sugar3/graphics/progresstoolbutton.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2016  Utkarsh Tiwari
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Contact information:
+# Utkarsh Tiwari    iamutkarshtiwari@gmail.com
+
+"""
+BETA
+"""
+
+import logging
+
+from gi.repository import Gtk
+from gi.repository import GObject
+
+from sugar3.graphics import style
+from sugar3.graphics import toolbutton
+from sugar3.graphics.toolbutton import ToolButton
+from sugar3.graphics.icon import Icon
+from sugar3.graphics.icon import get_surface
+from sugar3.graphics.palette import Palette, ToolInvoker
+
+
+class ProgressToolButton(ToolButton):
+    """
+    This class provides a simple wrapper based on the :class:`ToolButton`
+    Displays the progress filling the ToolButton icon.
+
+    Call update(progress) with the new progress to update the ToolButton icon.
+
+    The direction defaults to 'vertical', in which case the icon is
+    filled from bottom to top.  If direction is set to 'horizontal',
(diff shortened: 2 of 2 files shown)


quozl:
* added bug label, as it is for bug 4721,
* added feature label, since it adds a feature to the toolkit,
* reviewed.

iamutkarshtiwari:
@quozl  indentation fixed! :)

quozl:
You have a mix of spaces and tabs on the Makefile.am line, please change it to be consistent with the other lines in the file, then collapse (rebase) your patches and force push your branch.

iamutkarshtiwari:
Inconsistencies removed.

iamutkarshtiwari:
I think that I should rather add this file to the browse repository because it's only being used by the browse's webtoolbar. There is no use of adding a module to sugar-toolkit if it's only going to used by just one activity. @quozl What are your suggestions on this sir?

quozl:
It is more likely to be used by other activities and Sugar if it is in the toolkit, and I think it is a useful addition to the toolkit, so my preference is that it should be merged for 0.110.  To effect that, leave this pull request open waiting for our maintainer.

Meanwhile, for Browse, so that the next release may be used on older versions of Sugar such as 0.108, you should try to import the class from the toolkit, and if that fails use a local class in the Browse source repository.

quozl:
In testing, I've found that *set_tooltip*() must be called in order to update the progress on display.  In other words;

```python
    button.update(x)
    button.set_tooltip('something')
```

It would be better if the second call was not needed.  Have you any suggestions?

iamutkarshtiwari:
Do you mean that the update() function is dependent on set_tooltip() or do you need an alternate option to display the progress?

quozl:
update() seems to be dependent on set_tooltip(), in that the icon fill
does not change if only update() is called in a timeout callback.

At the same time, other timeout callbacks do cause updates.

Was discovered in development of the Clock activity pull request.

Workaround was to put extra set_tooltip() calls in.

iamutkarshtiwari:
Looking into this.

iamutkarshtiwari:
@quozl Try adding ```self.queue_draw()``` after ```242'nd line``` in the ```update()``` function. I think this should fix the problem. Let me know if it works, I'll update the patch.

quozl:
@iamutkarshtiwari, thanks, that fixes it.

For the record, https://github.com/godiard/clock-activity/pull/2 is the pull request related to my question.

samdroid-apps:
Awsome.  This is like the widget that we need.  If you're looking for more places to use it, maybe look at the file transfer notifications in sugar shell (to show download progress).

Vauge comments:

* API Docs are great.  Please do them.  See an example of the format with this patch:  https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/269/files
* Can I make this pulse as well?  With a progress bar, I would use the pulse to indicate that progress is being made, however I do not know any numbers.
* You duplicated lots of the methods of the class (eg. get_accelerator, do_clicked, ...), as well as some auxiliary functions (_add_accelerator, etc.) from the toolbutton.py file.  Can you import them and make the ProgressToolButton a subclass to inherent those methods?  Duplication is bad

Specific comments:

[https://gyazo.com/32bd9ad367023faef413f96c3139c89d](https://gyazo.com/32bd9ad367023faef413f96c3139c89d)

* Normal toolbuttons have a black background when the palette is down.  This is done through adding the "toolbutton" and palette-down css classes (just search for that in the toolbutton file).  Please add that feature.
* The sizing seems a little off, see the screenshot above.  That was tested when SUGAR_SCALING=100 (as in the environment variable)
* Please make the orientation value come from constants rather than random strings.  Eg. Gtk.Orientation.VERTICAL not 'verticacl'.  It is eaiser to see typos that way (AttributeError vs nothing)

Otherwise, this is amazing.

iamutkarshtiwari:
@samdroid-apps Updated and rebased commits.

Following defects are fixed->

- Documented the functions and the class.
- Functions duplication removed(apologies for that mistake). Inheritance implemented.
- Palette's background fixed.
- Orientation value now comes from constant (Gtk.Orientation.VERTICAL)

- The sizing of the download icon is initialized in ```webtoolbar.py to -> style.STANDARD_ICON_SIZE``` . It's pretty weird why it's appearing like that. I need to take a closer look but this shouldn't cause much issue.

quozl:
Tested on Fedora 18 with https://github.com/sugarlabs/browse-activity/pull/24.

Confirmed that background is black, and sizing looks right on an XO laptop.

iamutkarshtiwari:
@samdroid-apps I think it's good to go now..?

i5o:
wow!

I will test, but I think @samdroid-apps have the final decission

samdroid-apps:
This is good, but we are refactoring this to work better.  See the https://github.com/sugarlabs/browse-activity/pull/24 patch for the discussion.  I'll close this pull request until we work that out.

PR #311, opened 2016-03-07
adds a toolbutton to the bookmark tray

iamutkarshtiwari:
https://github.com/sugarlabs/browse-activity/pull/26

This PR adds a `toolbutton` to the tray module to be used by the `bookmark management feature` of Browse-activity

### DIFF
--- a/src/sugar3/graphics/tray.py
+++ b/src/sugar3/graphics/tray.py
@@ -175,6 +175,35 @@ class _TrayViewport(Gtk.Viewport):
             self.notify('can-scroll-next')
 
 
+class _TrayToolButton(ToolButton):
+
+    __gtype_name__ = 'SugarTrayToolButton'
+
+    __gsignals__ = {
+        'add-link': (GObject.SignalFlags.RUN_FIRST, None, ([]))
+    }
+
+    def __init__(self, icon_name):
+        ToolButton.__init__(self)
+        self._viewport = None
+
+        self.set_size_request(style.SMALL_ICON_SIZE, style.SMALL_ICON_SIZE)
+
+        self.icon = Icon(icon_name=icon_name,
+                         pixel_size=style.SMALL_ICON_SIZE)
+        # The alignment is a hack to work around Gtk.ToolButton code
+        # that sets the icon_size when the icon_widget is a Gtk.Image
+        alignment = Gtk.Alignment(xalign=0.5, yalign=0.5)
+        alignment.add(self.icon)
+        self.set_icon_widget(alignment)
+        alignment.show_all()
+
+        self.connect('clicked', self._clicked_cb)
+
+    def _clicked_cb(self, button):
+        self.emit('add-link')
+
+
 class _TrayScrollButton(ToolButton):
 
     __gtype_name__ = 'SugarTrayScrollButton'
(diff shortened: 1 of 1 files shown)


src/sugar3/graphics/tray.py
```diff
@@ -175,6 +175,35 @@ def _adjustment_changed_cb(self, adjustment):
...
+        # that sets the icon_size when the icon_widget is a Gtk.Image
+        alignment = Gtk.Alignment(xalign=0.5, yalign=0.5)
+        alignment.add(self.icon)
+        self.set_icon_widget(alignment)
+        alignment.show_all()
+
+        self.connect('clicked', self._clicked_cb)
+
+    def _clicked_cb(self, button):
+        self.emit('add-link')
```
samdroid-apps: This code is way to specific to browse activity

iamutkarshtiwari:
@samdroid-apps Could you please suggest an alternate way to implement this? Or the other way to implement this is create a new ```TrayToolButton``` module in browse-root directory ? What would you suggest?

quozl:
I don't understand how;
* the code is specific to the activity (it doesn't seem to be),
* the code being specific to the browse activity prevents the code from being generally useful for other activities in future.

Meanwhile the code in this pull request will end up being added to Browse activity repository anyway, otherwise Browse would not be supported on existing and older versions of Sugar.  But that's peripheral to the task of assessing the code for inclusion in the toolkit.

src/sugar3/graphics/tray.py
samdroid-apps: Can we rename the signal 'add-link' to something more generic, like 'activate'

src/sugar3/graphics/tray.py
```diff
@@ -261,6 +290,10 @@ def __init__(self, **kwargs):
         self._box.pack_start(self._viewport, True, True, 0)
         self._viewport.show()
 
+        self._add_bookmark = _TrayToolButton('list-add')
+        self._box.pack_start(self._add_bookmark, False, False, 0)
+        self._add_bookmark.show()
```
samdroid-apps: We probably shouldn't add this button to every tray - otherwise we need to update every activity that uses a tray to make use of the button.

What activities are using the tray button now?  How can we make an api that let's them add a traytoolbutton only if they want it?

icarito:
Well the tray as it exists now is pretty thin on features do we want a thicker sugar-toolkit or perhaps let Activities make their own trays? I echo sam's question, what other activities use this tray widget?

iamutkarshtiwari:
@icarito It is browse specific.

icarito:
@iamutkarshtiwari It resides in sugar-toolkit-gtk3 for some reason. Presumably because it was meant to be used by more activities. If none other than Browse use it perhaps it would be better to move it into browse?

iamutkarshtiwari:
@icarito I agree.

quozl:
@iamutkarshtiwari was this added to Browse?  Perhaps https://github.com/sugarlabs/browse-activity/commit/d25af13602b092f03ae8a1b94857e587c72b2531 ?

iamutkarshtiwari:
@quozl This PR was a part of this PR - https://github.com/sugarlabs/browse-activity/pull/26 which was closed.

quozl:
So you can close this now?

PR #312, opened 2016-03-26
Update speech.py

leonardcj:
Improve the i18n of speech strings with wiki links.  Please merge and then recreate the POT file.


samdroid-apps:
We had a mailing list discussion about this is believe.  Did we reach a positive consensus?  I don't know enough about i18n to comment.

godiard:
+1. Merged

leonardcj:
Thanks, the list discussion mentioned other possible improvements, these may come in future.  In addition, there will be a near future L10n landing.

PR #314, opened 2016-04-05
Set sugar cursor internally in sugar

erilyth:
Originally Gio.Settings were used to set the cursor. Set the cursor internally in Gdk so it doesn't effect the rest of the system.

Related PR in sugar: https://github.com/sugarlabs/sugar/pull/670

### DIFF
--- a/src/sugar3/activity/activity.py
+++ b/src/sugar3/activity/activity.py
@@ -385,6 +385,9 @@ class Activity(Window, Gtk.Container):
         self._bus = ActivityService(self)
         self._owns_file = False
 
+        sugar_cursor = Gdk.Cursor(Gdk.CursorType.LEFT_PTR)
+        self.get_window().set_cursor(sugar_cursor)
+
         share_scope = SCOPE_PRIVATE
 
         if handle.object_id:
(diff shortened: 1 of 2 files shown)


samdroid-apps:
This code only sets to the the left cursor from the current theme.  This patch does not fix the issue, what if the user has more than one cursor theme installed (Eg. Adwainta, HighColor - all pulled in by GNOME).  This patch doesn't like do anything!

erilyth:
@samdroid-apps , I wanted to create the cursor using a pixbuf, which I could put in this place, so the sugar cursor is drawn. How can I get the pixbuf of the cursor from the sugar-artwork of the sugar cursor? I tried if I could create an Icon of the cursor and get the pixbuf from there but was unsuccessful.

erilyth:
@samdroid-apps , Ah I understood what is happening, my default cursor on my system is the one that sugar uses so its getting set to that when I use this patch.
Is there any way I can get a pixbuf of the cursor from the sugar-artwork/cursor, so that the cursor can be loaded using the pixbuf? I think that this needs to be done everywhere where we are currently just setting the cursor using Gdk.Cursor(Gdk.CursorType(xyz)). Not just for the standard cursor, but for all the different types of cursors.

quozl:
@erilyth, I agree that the current code using Gdk.Cursor in many places is very messy; it could be refactored into a common function.  Some of my cursor fixes in the past year have been made with the intention of eventually refactoring.  You might consider them.

PR #315, opened 2016-04-17
Allow to run activities without Sugar shell.

icarito:
- Handle lack of GSettings gracefully.
- Still requires sugar-datastore.
- Supports avoiding X11 docks/panels.
- Provides icons for Activity windows.

Try it outside Sugar. Go to an Activity directory and run 'sugar-activity'.
Tested it with Terminal, Finance, Write, Browse, Memorize under XFCE4.
Tested in Gnome under OLPC-OS.
Also works from Sugar Terminal Activity.
Does not affect regular Sugar operation.

### DIFF
--- a/bin/sugar-activity
+++ b/bin/sugar-activity
@@ -35,8 +35,10 @@ DBusGMainLoop(set_as_default=True)
 from sugar3.activity import activityhandle
 from sugar3 import config
 from sugar3.bundle.activitybundle import ActivityBundle
+from sugar3.activity.activityfactory import get_environment
 from sugar3 import logger
 
+from sugar3.bundle.bundle import MalformedBundleException
 
 def create_activity_instance(constructor, handle):
     activity = constructor(handle)
@@ -69,7 +71,13 @@ class SingleProcess(dbus.service.Object):
 
 
 def main():
-    parser = OptionParser()
+    usage = 'usage: %prog [options] [activity dir] [python class]'
+    epilog = 'If you are running from a directory containing an Activity, ' \
+             'the argument may be ommitted. Otherwise please provide either '\
+             'a directory containing a Sugar Activity [activity dir], a '\
+             '[python_class], or both.'
+
+    parser = OptionParser(usage=usage, epilog=epilog)
     parser.add_option('-b', '--bundle-id', dest='bundle_id',
                       help='identifier of the activity bundle')
     parser.add_option('-a', '--activity-id', dest='activity_id',
@@ -89,18 +97,34 @@ def main():
 
     logger.start()
 
-    if 'SUGAR_BUNDLE_PATH' not in os.environ:
-        print 'SUGAR_BUNDLE_PATH is not defined in the environment.'
-        sys.exit(1)
+    activity_class = None
+    if len(args) == 2:
+        activity_class = args[1]
+        os.chdir(args[0])
+    elif len(args) == 1:
+        if os.path.isdir(args[0]):
+            os.chdir(args[0])
+        else:
+            activity_class = args[0]
 
-    if len(args) == 0:
-        print 'A python class must be specified as first argument.'
-        sys.exit(1)
+    os.environ['SUGAR_BUNDLE_PATH'] = os.path.abspath(os.curdir)
 
     bundle_path = os.environ['SUGAR_BUNDLE_PATH']
     sys.path.append(bundle_path)
 
-    bundle = ActivityBundle(bundle_path)
+    try:
+        bundle = ActivityBundle(bundle_path)
+    except MalformedBundleException:
+        parser.print_help()
+        exit(0)
+
+    if not activity_class:
+        if bundle.get_command().startswith('sugar-activity'):
+            print 'Guessing python class from activity.info!'
+            activity_class = bundle.get_command().split(" ")[1]
+
+    if 'SUGAR_VERSION' not in os.environ:
+        environ = get_environment(bundle)
 
     os.environ['SUGAR_BUNDLE_ID'] = bundle.get_bundle_id()
     os.environ['SUGAR_BUNDLE_NAME'] = bundle.get_name()
@@ -114,7 +138,7 @@ def main():
     gettext.bindtextdomain('sugar-toolkit-gtk3', config.locale_path)
     gettext.textdomain(bundle.get_bundle_id())
 
-    splitted_module = args[0].rsplit('.', 1)
+    splitted_module = activity_class.rsplit('.', 1)
     module_name = splitted_module[0]
     class_name = splitted_module[1]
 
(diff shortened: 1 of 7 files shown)


icarito:
Also tested running activities from GNOME under OLPC-OS.

icarito:
Here's a feature page:
https://wiki.sugarlabs.org/go/Features/FreeDesktop_integration

icarito:
The intention is to support this experiment. Might even include in patch?
https://github.com/icarito/sugar-launcher-applet

quozl:
Reviewed c85738d.  Like it.

Suggest packaged activities might also provide .desktop files.

Suggest sugar-activity might also accept path to unpacked bundle.

godiard:
Very nice. 
A few suggestions:
* You can avoid code duplication using the functions get_color(), get_nick_name() in the other places where the settings are read.  
* Remove commented code.

leonardcj:
It would be nice if the changes to the POT for sugar-toolkit-gtk3 could be incorporated in this pull request, please.

bin/sugar-activity
```diff
@@ -89,19 +89,31 @@ def main():
...
-        sys.exit(1)
+    os.environ['SUGAR_BUNDLE_PATH'] = os.path.abspath(os.curdir)
 
     bundle_path = os.environ['SUGAR_BUNDLE_PATH']
     sys.path.append(bundle_path)
 
     bundle = ActivityBundle(bundle_path)
 
+    if len(args) == 0:
+        if bundle.get_command().startswith('sugar-activity'):
```
samdroid-apps: Couldn't the launcher just pass this info?

bin/sugar-activity
```diff
@@ -89,19 +89,31 @@ def main():
...
+            print 'Guessing python class!'
+            command = bundle.get_command().split(" ")[1]
+        else:
+            print 'A python class must be specified as first argument ' + \
+                    'or else run it inside a python gtk3 activity directory.'
+            sys.exit(1)
+    else:
+        command = args[0]
+
+    if 'SUGAR_VERSION' not in os.environ:
```
samdroid-apps: Mate, your launcher really should put sugar version in the environment.  Browse needs to use it, for example.

src/sugar3/activity/activity.py
```diff
@@ -436,13 +436,23 @@ def __init__(self, handle, create_jobject=True):
                                            self.__jobject_updated_cb)
         self.set_title(self._jobject.metadata['title'])
 
+        if 'SUGAR_VERSION' not in os.environ:
```
samdroid-apps: Why don't we always set the icon?  It will do no harm under sugar right?

src/sugar3/activity/activity.py
```diff
@@ -436,13 +436,23 @@ def __init__(self, handle, create_jobject=True):
...
+
     def run_main_loop(self):
         Gtk.main()
 
     def _initialize_journal_object(self):
         title = _('%s Activity') % get_bundle_name()
-        settings = Gio.Settings('org.sugarlabs.user')
-        icon_color = settings.get_string('color')
+       
+        if 'org.sugarlabs.user' in Gio.Settings.list_schemas():
```
samdroid-apps: Couldn't the launcher just set up the gsettings properly?  This feels very dirty - many things expect the gsettings to work.

src/sugar3/activity/activity.py
```diff
@@ -1251,7 +1262,16 @@ def get_activity_root():
     if os.environ.get('SUGAR_ACTIVITY_ROOT'):
         return os.environ['SUGAR_ACTIVITY_ROOT']
     else:
-        raise RuntimeError('No SUGAR_ACTIVITY_ROOT set.')
```
samdroid-apps: Maybe the launcher could set the activity root as well?

Also, imports should be at the top of the file?

samdroid-apps:
Hi @icarito, thanks for the patch.

An important consideration for this patch is code architecture.  I think that right now, it is not a good architecture.  It is hacking around things that the toolkit (and also activities) previosuly assumed were present; eg. compensating for the lack of gsettings.

I think that maybe we could do this better by having a "launcher script" that setup the environment for the activity (eg. ensure that the gsettings are set up properly), and then launch the activity as per the activity.info description.  That script could be shipped with sugar, or sugar-toolkit-gtk3.

Maybe the script could look like:  `sugar-external-launch-activity /usr/sugar/activities/Browse.activity`?

quozl:
I'm fine with this level of hacking around things at this stage of feature development; none of the hacks should cause problems in this release.  Architectural improvements can come later.

We already have `sugar-launch` so perhaps it could be adjusted to recognise that the process environment is not Sugar but some other desktop.  That would let us use the same tool in multiple places.

samdroid-apps:
What is the need to "handle the lack of gsettings"?  If the user has installed sugar, won't they have the gsettings schemas installed and therefore available?  This is what happened when I tested on my Fedora machine.

quozl:
> What is the need to "handle the lack of gsettings"?

org.sugarlabs.user may be either not set, or not available.

> If the user has installed sugar, won't they have the gsettings schemas installed and therefore available?

Not if they have only installed the toolkit.

Not if the GSettings infrastructure is unavailable, as happens on some desktop environments that predate or are not subsets of Gnome.

icarito:
Everybuddy thanks for your kind reviews. I think I've handled most everybuddy's concerns with the latest version of the patch. Of course more could be done. Looking forward.

samdroid-apps:
Love it.  Looking good.

Just 1 thing before the merge.  Please move the import statements to the top of the file, with the other import statements.

icarito:
Done!

icarito:
I imagine sugar-toolkit-gtk3 might be uploaded to the python cheese shop :-)

icarito:
Hi this decouples sugar-toolkit-gtk3 from the sugar-shell which might be a first step towards running under broadwayd (HTML5). @godiard @quozl are we good enough to merge? Thanks in advance!

icarito:
> Suggest packaged activities might also provide .desktop files.

Do you mean activities themselves should provide .desktop files? Perhaps we could provide a script `activity_info2desktopfile`.

quozl:
@icarito,
* i've cherry-picked 2f2b2d20d8ba8b6b94941dc0fed06a82c0f41a9f into my OLPC branch, thanks, so it's good enough for me, no idea what Sugar Labs thinks,
* desktop files in the activity packages would let other desktops start activities, and the bundlebuilder could be improved to make desktop files.

samdroid-apps:
Na, this stuff is all good.  @quozl, feel free to do the merge in future.  I will actually merge it this time.

icarito:
@samdroid-apps moving imports to the top of the file broke GTK2 activities! I've provided #317 to fix it, or we could move the import to the conditional in this case.

icarito:
(credit goes to @quozl for figuring the issue out!)

icarito:
@leonardcj I had answered your concern by updating the commit message but I guess github doesn't email updated commit logs:

    It would be nice if the changes to the POT for sugar-toolkit-gtk3
    could be incorporated in this pull request, please.
        - There were no changes to POT files as part of this patch. Maybe
        it is worth translating low level command line tools, not sure. Currently they aren't, and changing that goes beyond the scope of this patch.

icarito:
Please also review and merge the companion patch to sugar-toolkit:
https://github.com/sugarlabs/sugar-toolkit/pull/3

Thanks in advance!

PR #317, opened 2016-05-17
Fix regression for GTK2 activities caused by 2f2b2d2

icarito:
The issue was caused by importing from activityfactory which caused GObject to be imported. This conflicts with GTK2. (Reported and original patch by James Cameron)
We should create the sugar profile's activity root if it doesn't exist yet, when running outside Sugar. But activity factory can't be used, because of this incompatibility.


icarito:
Added minor fix, got activity_id wrong the first time.

samdroid-apps:
Reviwed, not tested.  Please merge.

PR #321, opened 2016-06-05
Generate .desktop files for activities during install

samdroid-apps:
Do you want to review this @icarito?

This will allow .desktops for all of the activities to be
packaged.  In the future, we could ship a special launcher script
so that the user can choose the jobject to run, but for now
we just open the activity using the command specified in the
activity.info file.

To test, just `./osbuild shell`, `build` and then `python setup.py install` an activity.  You will need to copy the desktop file, something like:

    cp ~/sugar-build/build/out/sandbox/install/share/applications/org.laptop.WebActivity.activity.desktop  /usr/local/share/applications/

Alternatively, don't use osbuild and don't do the extra copy!

------

*Using the issue tracker as a mailing list*

I'm really excited about making all the `.desktop` entries, because it will let us use AppStream!  I was so excited that I wrote a patch to send to AppStream [1] after we merge something like this.  AppStream is the cool metadata collecter that powers gnome-software (and all the other modern app stores).  It would be really slick to intergrate an AppStream based interface in Sugar, and having activity.info metadata natively intergrated in that would be really cool.  Kind of like ASLO2v2, but using GNOME techologies rather than reinventing the wheel and using html.

Here is what the appstream patch does to rpm files.  It gets the metadata from activity.info, but it will only make sense if we also make the .desktop files too (because otherwise GNOME software users will be confused, and probably hate our patch):

screenshot from 2016-06-05 21-03-54

[1]  https://github.com/hughsie/appstream-glib/compare/master...samdroid-apps:sam-master?expand=1

### DIFF
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -31,7 +31,9 @@ import subprocess
 import re
 import gettext
 import logging
+from glob import glob
 from fnmatch import fnmatch
+from ConfigParser import ConfigParser
 
 from sugar3 import env
 from sugar3.bundle.activitybundle import ActivityBundle
@@ -255,7 +257,7 @@ class Installer(Packager):
         Packager.__init__(self, builder.config)
         self.builder = builder
 
-    def install(self, prefix, install_mime=True):
+    def install(self, prefix, install_mime=True, install_desktop_file=True):
         self.builder.build()
 
         activity_path = os.path.join(prefix, 'share', 'sugar', 'activities',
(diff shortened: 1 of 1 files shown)


icarito:
Great! The code looks fine! I am not using sugar-build but instead have installed locally using `make install`. 

I would've expected to find them under `/usr/local/share/applications/` dir. I guess that's an issue with bundlebuilder not using `/usr/local` as default prefix?

I first tried it with GTK2 activities, I guess it needs to be ported for those too!

Happy to see this development! +1 for merging from me.

icarito:
Just a small comment from @kaametza - please add `Categories=Education;` to the Desktop files.

samdroid-apps:
>  I would've expected to find them under /usr/local/share/applications/ dir. I guess that's an issue with bundlebuilder not using /usr/local as default prefix?

The bundlebuilder defaults to the `/usr` prefix by default.  That is a bit funny, but you can use the `--prefix` option to change that.  My patch is consistent with the prefix set, so it shouldn't introduce a new issue.

For GTK2 activities, they could update the bundlebuilder (in setup.py) to the sugar3 copy.  I think you can update the bundlebuilder version separate from the Gtk+ version.

I'll add the Education category asap.

davelab6:
THIS IS AWESOME!!! :D I agree about the ASLOv2 comment, this is great :)

samdroid-apps:
@icarto, I force pushed with the changes to add the Education category.  Please review (if needed) and merge!

davelab6:
FYI https://eischmann.wordpress.com/2016/06/06/flatpak-is-gaining-momentum/

icarito:
Cool! But I still don't see the Education category, did you commit before force pushing?

samdroid-apps:
@icarito, good catch!  I updated it again.

PR #323, opened 2016-06-13
Popup window size_set() bugfix

iamutkarshtiwari:
The value assinging technique (size tuple) is now fixed.


samdroid-apps:
Hi, you just pass a tuple like:

    w.set_size((100, 200))

This allows you to do things like:

    w.set_size(PopWindow.FULLSCREEN)

It also allows the GObject property:

    w.props.size = PopWindow.HALF_WIDTH

Is the tuple usage causing you a bug?

PR #326, opened 2016-07-03
Unify palette popdowns

samdroid-apps:
Previously there was the concept of having 2 popdowns for a palette.
The first popdown showed the of tooltip the palette, and the 2nd
showed the content of the palette.  However, this was confusing
to new users, and made users wait for little reason.

This commit removes the 2nd popdown.  An example of this is when you
mouse over an icon to invoke the palette.  Previously, it showed
the tooltip, then made you wait, then showed the content of the
palette.  With this patch, the whole palette is shown at the
first popdown, making a snappy UX.

See https://www.sam.today/blog/sugar-with-instant-palettes/ for more background

Mailing list reference:  http://lists.sugarlabs.org/archive/sugar-devel/2016-May/052808.html

### DIFF
--- a/src/sugar3/graphics/palette.py
+++ b/src/sugar3/graphics/palette.py
@@ -123,7 +123,6 @@ class Palette(PaletteWindow):
         self._secondary_text = None
         self._icon = None
         self._icon_visible = True
-        self._palette_state = self.PRIMARY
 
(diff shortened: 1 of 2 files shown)


davelab6:
I think this is great :)

PR #327, opened 2016-07-04
Save As Popup

iamutkarshtiwari:
'Save As' popup added which appears on activity close. It allows user to provide instance name to the user.  - https://www.youtube.com/watch?v=xcvBH7zzFBo
desktop-animation


iamutkarshtiwari:
The patch needs clean up. This is just a beta version.

icarito:
Allright I just tested this feature.
My comments:
* [ ] Text entry is way too small.
* [ ] Text entry does not offer current object title for changing
* [ ] Save as/Discard button is ambiguous (and it takes a lot of time to act for some reason).
* [ ] "Same name" warning is ambiguous. The same name object is the object itself.
* [ ] Overwrite action should happen if user insists with stop button.
* [ ] There should be a way to cancel the alert.
* [ ] It should not be possible to enter an empty name.

Some serious thought needs to go into this interaction, as it is a very important one. 

We must take into consideration that Sugar *already* saves the object if the user changed activities or views before, even if it was the first instance (of object creation).

Perhaps the alert could work this way: It could have default actions for each question. Pressing Stop button again will simply take default actions (I instinctively tried this, i.e. telling the system I really want to stop, don't ask me questions).

Take this as an idea:

* Confirm saving as: [Object title] - ***save***, **cancel**
  * If a user changes the object title, the buttons could switch to ***rename***, **save copy**, **cancel**
  * In addition, a menu button could be added for the following secondary actions: **discard**, **save to...***

samdroid-apps:
I agree with @icarito about the issues with the current implementation.

Here are a few of my takes on it:

* What does the disregard button do?  Is it suppose to delete the entry?  Isn't that what was discussed on the mailing list?  When I tested, the disregard button just did nothing and closed the activity :(
* Disregard is such a fancy word.  Could we use more direct words like "save" and "delete" rather than "confirm" and "disregard"?
* Can it be a popwindow?  Then we get the full width of the screen, and you can put the full journal details view there - making it easy to add a description.  Using the full details view (rather than making a new view) is nice from a consistency perspective - it is a view that users are familiar with.
* Can the options be "save" and "delete".  Save will save the journal object, with whatever the title and metadata the user has set.

iamutkarshtiwari:
@icarito @samdroid-apps I have fixed some of the issues. Please test the newest commit.

@samdroid-apps When the user selects the ```discard``` option, the activity data (for eg. text in case of Write activity) doesn't get saved on close. Only the object entry appears in datastore for statistical purpose but not the activity data.

samdroid-apps:
> @samdroid-apps When the user selects the discard option, the activity data (for eg. text in case of Write activity) doesn't get saved on close. Only the object entry appears in datastore for statistical purpose but not the activity data.

I thought that one of the main reasons to add this change was the delete journal clutter by letting the user delete entries more quickly.

What is the use case of saving the metadata but not the entry data?

iamutkarshtiwari:
@samdroid-apps Mr. Anderson asked me to build it that way :/

icarito:
"When the user selects the discard option, the activity data (for eg. text in case of Write activity) doesn't get saved on close."

But it's already saved before close! You mean it gets deleted?

iamutkarshtiwari:
@icarito Yes. When the activity is open, a duplicate copy is created. If the user selects to 'discard' the data, the duplicate copy is deleted without saving the data. And if the user selects to save the data, the duplicate copy is retained by the new name provided by the user in the text entry and the data the is saved to that dupicate instance without overwriting the original instance.

icarito:
Here's comments by Tony on private thread:
    "Consider:

You open Paint and create a red X and save it as x. Then you open x and change the image to a blue O. You want to save it as O, so you change the name.
Now try to open X. It is gone.

Suppose you open Paint to show someone how to select a brush and give it a color. You scribble something. When asked for a name, you click on 'quit' because there is not need to save this image.

The alert should be simple:

Please give your project a title.   [Untitled]     Save   Quit  

The default title is 'Untitled' not 'Write.activity'. If the user clicks save with the title unchanged, it is ignored. If the user clicks on Quit, the activity quits (with no save of the Document - but the metadata is saved). if the user resumes an activity the alert shows:

Change the title to save as a new document    [X]  Save Quit

 The user changes the title [X] to [Y]. Then Save saves Y to the Journal and X is still in the Journal as well. If the user doesn't change and clicks Save, the X object is deleted and the new X version is saved. if the user clicks on Quit - the document is not saved, the original X document is in the Journal
and the activity quits."

I can't find Tony's github user to mantion him.

My comment is:<br>
Clicking Stop, and then "Quit" does not tell me as a user that my work will be deleted (this is what you seem to be suggesting).

"Change the title to save as a new document": seems long to me. Also it's not always a document (e.g. Browse session). 

 if the user clicks on Quit - the document is not saved, the original X document is in the Journal
and the activity quits." - but the document may already have been saved (on view change, for instance).

icarito:
"If the user selects to 'discard' the data, the duplicate copy is deleted without saving the data. "

This is the part that I don't understand, Sugar will save the data on every view change. So you mean it will delete the duplicate (that has been previously already saved).

icarito:
I tried the new version and it works much better. I'm not convinced by the text in the alert. You seem to imply two objects can't have the same name, but they can. Objects are identified by a unique hash id in the journal.

Save as / Quit is a strange button label.

One of Sugar's design principles is simplicity.

src/sugar3/activity/activity.py
```diff
@@ -1069,12 +1100,81 @@ def _keep_failed_dialog_response_cb(self, alert, response_id):
...
 
     def can_close(self):
-        '''
-        Activities should override this function if they want to perform
-        extra checks before actually closing.
-        '''
+        """Activities should override this function if they want to perform
+        extra checks before actually closing."""
 
-        return True
```
icarito: After your patch I can't close ImageViewer or Chat. This is likely why.

src/sugar3/activity/activity.py
```diff
@@ -1212,7 +1317,7 @@ def __get_filters_cb(self):
         }
         filter_dict = dbus.Dictionary(filters, signature='sv')
         logging.debug('__get_filters_cb %r' % dbus.Array([filter_dict],
-                      signature='a{sv}'))
+                                                         signature='a{sv}'))
```
icarito: If you have cosmetic changes make them on another PR. It makes it hard to see what you changed.

src/sugar3/activity/activity.py
```diff
@@ -1069,12 +1100,81 @@ def _keep_failed_dialog_response_cb(self, alert, response_id):
...
+        self.remove_alert(alert)
+
+        # 'Save as' option
+        if response_id == Gtk.ResponseType.OK:
+            self._show_saveas_alert()
+
+        # Overwrite option
+        if response_id == Gtk.ResponseType.CANCEL:
+            self._skip_save_datafiles = False
+            datastore.delete(self._jobject.get_object_id())
```
icarito: When choosing to overwrite do you delete the previous copy or do you always keep a copy?

icarito:
Currently my main concern from a UI pov is that choosing "Stop" and then "Quit" will result in the user discarding his work. This is not intuitive to me.

icarito:
"Mr. Anderson asked me to build it that way." - That's a bad reason, you should believe in what you do! There are better ways to collect metadata than to clutter the journal. This has been a bad use of it.

src/sugar3/activity/activity.py
iamutkarshtiwari: The reason 'ImageViewer' and 'Chat' didn't close is because these two activities override the 'can_close()' funtion of 'activitly.py' which is responsible to display my alert. 

I have figured out a solution and commit a new tweak for the same.

src/sugar3/activity/activity.py
iamutkarshtiwari: I have an 'AutoPep8Fix' package in my sublime which fixes the Pep8 errors on 'Ctrl+S'. I'll make sure to disable it.

src/sugar3/activity/activity.py
iamutkarshtiwari: On choosing overwrite the data gets saved onto the original instance whereas the duplicate gets deleted.

iamutkarshtiwari:
@icarito I agree. But the 'SaveAs' feature was proposed by Mr. Anderson and I'll have to abide by the requirements he specifies. I can provide him with my personal opinions/suggestions but I can't do nothing unless he approves the patches.

tony37:
This thread would be a lot more useful if you took the time to understand Sugar as well as this PR.

"Mr. Anderson asked me to build it that way." - That's a bad reason, you should believe in what you do! There are better ways to collect metadata than to clutter the journal. This has been a bad use of it.

This is crazy. I am mentoring this feature. Granted Utkarsh can proceed as he feels best but we don't need to involve him in these quibbles.

Any Journal object consists of two elements: a metadata file and an (optional) data file. The data file normally conforms to a meta_type and is from the point of view of the user, a document. 

The metadata needs to be saved in every case. The document only if the user chooses. Note: that any user is free to delete any Journal object at any time without getting permission from anyone.

tony37:
Currently my main concern from a UI pov is that choosing "Stop" and then "Quit" will result in the user discarding his work. This is not intuitive to me.

This is why you should discuss the issue. The alert offers two options: save and quit. The alert appears because the user has clicked on 'stop' to request terminating the activity.

The current activity.py does indeed save the document without action by the user (we used to have 'keep' which gave the user the option to save and then continue. This was removed.

However, the implementation is flawed. The jobject used by activity.py in case of resume is the original jobject_id. This means, unknown to the user, he is making irretrievable changes to his document. Hence, 
the scenario quoted by Utkarsh. 

Utkarsh makes a copy of the original jobject and this is what is changed. This enables the save option after a name change, to keep both the original document and the new instance as a user should expect.

tony37:
When choosing to overwrite do you delete the previous copy or do you always keep a copy?

If you take the trouble to read the description of the feature, you will know that if the user resumes an activity with a title and chooses not to change the title, then the modified document is saved and the original is deleted. This is exactly what happens now. If the user changes the name, then the new document is saved under its new name and the unmodified original is also in the Journal.

tony37:
If you have cosmetic changes make them on another PR. It makes it hard to see what you changed

Who decides what changes are important and which are 'cosmetic'. This is the PR.

tony37:
The reason 'ImageViewer' and 'Chat' didn't close is because these two activities override the 'can_close()' funtion of 'activitly.py' which is responsible to display my alert.

I have figured out a solution and commit a new tweak for the same.

Utkarsh: 'can_close' is a function which can be supplied by the activity. In that case, the copy in activity.py is not called. What you need to do is post the alert after can_close. This way from the user's view, the alert will be seen before close. When the user responds to the alert, the activity can close.

tony37:
I tried the new version and it works much better. I'm not convinced by the text in the alert. You seem to imply two objects can't have the same name, but they can. Objects are identified by a unique hash id in the journal.

From the user perspective, documents should have a unique name. Granted in standard software, different files may have the same name in different directories. We only have the Journal. 

Save as / Quit is a strange button label.

The button options should be 'save' and 'quit'. The text on the button may differ from the start new case when the title is 'Untitled' or 'Write.activity' - perhaps: Please give a name to your project. In the resume case, the text should refer to the option to save (no change) or save as (change). Perhaps, Change the name of your project to save a new document. 

One of Sugar's design principles is simplicity.

iamutkarshtiwari:
@icarito @tony37  The 'Chat' and 'ImageViewer' deadlock on close issue resolved in the latest commit. Please test it and share your suggestions

@tony37 We all together can workout this feature's design and functionality :) After all, our contributions are going to benefit our Sugar community.

src/sugar3/activity/activity.py
davelab6: (FWIW I don't mind including such cosmetic changes with Python, because the language has whitespace requirements, and nonconformance with pep8 is unwise :)

samdroid-apps: ~~Cosmetic changes are never rejected~~

Flake8 fixes are always accepted.  But better to keep them in a separate commit (eg. "Flake 8 fixes)

icarito:
As promised here's a screenshot of Sugar 0.92's "save as" dialog:
captura de pantalla de 2016-07-14 21-33-17

As per our meeting today I think Martin Dengler's proposal to land the user in the Journal after quitting with focus on Object Title would be a good solution.

icarito:
Correction screenshot is from Sugar 0.90.2

tony37:
As you might expect, I totally disagree.

First, by that time the damage has been done. Either the 'save as' 
option has been taken with a new object created or the 'resumed' 
document has been overwritten - as at present.

This feature has two objectives:

     1 - require Journal items to have a name supplied by user
     2 - enable the user to choose whether to save the object as a new 
one or to simply save the modified document.

This proposal addresses neither objective.

You are not going to change your point of view so the only solution I 
can see is to provide a gsetting option so that users and deployments 
can decide on how they want this to work.

Re: the screenshot. Help me - as I remember this dialog was modal and 
required an entry in the description area - leaving the title as 'Write 
activity' was ok.
This was to implement Walter's desire that the description serve as the 
message does in git.

Tony

tony37:
I think user testing can resolve such ux disagreements.

A setting to choose from both is not a wise design decision because there
is still a question of which is default and how easy it is to set.

Since Tony is advocating for a change, perhaps you could do a user test.
Have you done one before? :)

samdroid-apps:
> This feature has two objectives:
>
>     1 - require Journal items to have a name supplied by user
>     2 - enable the user to choose whether to save the object as a new
one or to simply save the modified document.
>
> This proposal addresses neither objective.

I think that user testing is invaluable there.  But I think that @icarito very clearly addresses item number 1 - the proposal literally puts the user in front of a screen dedicated to managing activity metadata.

For addressing proposal 2, did you want to make https://github.com/sugarlabs/sugar/pull/699 false by default?  I would support making that false by default, now that I understand the rational.

But seriously, if you do some user testing and write up some scenarios for the users, I can also find some other people and carry out the same tests too!  More testing the better!

tony37:
Sam

I don't understand you. #327 satisfies both objectives.

"The proposal literally puts the user in front of a screen dedicated to 
managing activity metadata." Clearly, this must be something completely 
different from $327 or #699. Perhaps you could describe this proposal in 
more detail. Does it have a PR? Could you be referring to Sebastian's 
suggestion to switch the user to the Journal view on close?

#699 is a completely different feature. Yes, it should not change Sugar 
unless the user or deployment chooses. While I many deployments have 
asked for this capability, it is not consistent with Walter's decision 
to provide resume from the home view as the default behavior for Sugar.

I have no idea how to test the first item you describe - it has nothing 
to do with either #327 or #699. I have had 'resume' implemented at least 
5 schools with nearly 500 XOs operating Sugar. I am not sure what more 
user experience is needed. Remember, this behavior was the way Sugar 
worked through its intial releases and may, for all we know, be the way 
it still works on the majority of operational XOs.

User tests on #327 seem unnecessary, We all use it everyday on Linux, 
Windows, or Mac. My most common use is on nano where when I close a 
buffer, I must give it a name. If it has a name, I am given the 
opportunity to change it. Virtually all document-handling applications 
that I am aware of work this way.

Tony

iamutkarshtiwari:
Here are the latest changes - 
desktop-animation

davelab6:
Love that GIF :)

Tony, I think user testing each change is always a good idea, if only to
validate that your assumption the change is for the better is true :)

tony37:
Hi, Dave

We can agree to disagree. In the first place, as Sam Parkinson found 
out, finding users representative of our target users is hard. Children 
in the Western world have experience with computers and smart phones and 
so on.

Second, I am unaware of user testing for most of the changes made to 
Sugar. For example a check box was added to the Journal View. How was 
that user tested?

We have trouble even getting changes documented for our users which is 
certainly required.

Tony

samdroid-apps:
> Hi, Dave
> 
> We can agree to disagree. In the first place, as Sam Parkinson found
> out, finding users representative of our target users is hard. Children
> in the Western world have experience with computers and smart phones and
> so on.

I never remember finding that out.

I want Sugar to be a great piece of software - one that even kids with experience of other computers and smart phones would *choose*.

I thought that we wanted to have an inclusive target market, of both kids who are the first computer owners in their family, and kids who are [Born Mobile](https://www.youtube.com/watch?v=p1w2IxMWQ1Y).

> 
> Second, I am unaware of user testing for most of the changes made to
> Sugar. For example a check box was added to the Journal View. How was
> that user tested?

Good, let's change that!  Let's start at the Open Source Usability blog:  http://opensource-usability.blogspot.com.au/2012/11/how-many-testers-do-you-need.html

tony37:
Hi, Sam

Let's start with the project view in the Journal.

As I remember, you reported failure in your user test of 'on-boarding'. 
As I also recall, the users tested felt that the 'on-board' feature held 
them up because they already knew how to do it. This is unlikely to be 
the reaction of our deployed users who have much more limited computer 
experience. If you want to test Sugar features with kids 'born mobile' 
great. This is particularly important if the question you are asking is 
would this feature attract new Sugar users from the Western world who 
are blessed with access to computers and to high-speed internet 24/7.

Tony

iamutkarshtiwari:
I have generated a similar PR for ```sugar-toolkit``` too. https://github.com/sugarlabs/sugar-toolkit/pull/4

iamutkarshtiwari:
@samdroid-apps @icarito @davelab6 Please test this PR and share your reviews. I have simplified the popup's UI as suggested.

quozl:
Reviewed.  I'm fine with this if the new behaviour can be disabled using `gsettings`.  Several deployments who work with OLPC won't be interested in returning to the 0.92 save-as dialog days, so I'll need a way to turn it off.

i5o:
Added to 0.112 milestone; we should re-discuss this.

tony37:
It works. I am installing in 13.2.8 for use in Rwanda. Adding a gsetting should be simple, This  has already been done for the 'resume-activity' feature which has been enabled by setting resume-activity  to false. The screen shot feature PR675 should also be implemented with a gsetting control.

What needs to be discussed is how to expose these and other gsetting options to users and deployments. Perhaps something like about:config.

i5o:
The question is, will it be enabled as default?

On 1/2/17, Utkarsh Tiwari <notifications@github.com> wrote:
> @i5o @tony37 This patch is working fine (tested well). I agree with Mr.
> Anderson. We can add a gsetting for enabling/disabling this feature in
> Sugar, something like -
> ```gsettings set org.sugarlabs.user save-as true ```
>
> @davelab6 @quozl @samdroid-apps What do you think?
>
> --
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/327#issuecomment-269945857


-- 
Ignacio Rodríguez

iamutkarshtiwari:
@i5o Let's set it to ```disabled``` by default. Should any user be interested in exploring the feature, they can set it ```enabled``` from gsettings option

quozl:
If nobody is willing to drive this feature to completion following the consensus above, then I'd say momentum is lost and it would be better to close the pull request until there is interest again.  I'll close it in a few days if there is no material progress.

tony37:
What's to complete. The request is to merge the Pull Request. The only apparent open issue is whether this feature should be enabled by a gsetting with the default disabled. 

That seems the correct strategy preserving compatibility for the user unless they know what they want.

While the above statement is that there is no conflict, installation on 0.110 results in several activities such as TurtleBlocks crashing. This needs to be investigated, of course.

I do not understand the comment  about loss of momentum. This PR was developed as part of GSOC16. The work was completed and presented for merger. We are awaiting action by the responsible developers.

tony37:
I am unfamiliar with the technical term 'momentum'. The request is 
clear, merge the changes with the master. This PR was developed as part 
of GSOC16 so any momentum, whatever that is, is the responsibility of 
the developers.

The material progress needed is to test the PR with the current 
repository since it is presented as a patch against an unspecified 
commit. The action is waiting for the responsible developers to act on 
the request to merge this PR and has been for months.

It is proposed that this feature be made disabled by default and to be 
enabled by a gsetting (as was done with the resume PR). This is 
appropriate preserving current behavior but making it available to those 
that want it.

Tony

quozl:
Waiting for the pull request to;

- [ ] add a gsetting, with a default,
- [ ] use the gsetting,
- [ ] resolve reported incompatibility with TurtleBlocks,

And since I'm quite surprised that it had an impact of activities, I'd also suggest;

- [ ] evidence of testing with other activities.

tony37:
Hi, James

This PR worked with 0.108. My belief is that changes to the files 
involved after 0.108 conflicted with the patches. Since two of the files 
affected are sugar/activity.py and sugar3/activity.py, an impact on 
activities of a faulty merge seem inevitable.

Since some unknown person is responsible to merge PRs, presumably that 
person knows how to tell when the merge conflicts. Apparently there is 
nothing that the developer of a PR can do about this.

Tony

quozl:
Developer of a pull request is jointly responsible with all other developers to resolve bugs and merge conflicts.  There is no merge conflict at the moment between this pull request and the master branch.

We're working toward 0.112 now, with 0.110 released some time ago, and 0.108 now quite old; so that it worked with 0.108 and TurtleBlocks is of diagnostic use to whomever will be looking into the bug.

I've no change to what I'm waiting for in my reply above.

tony37:
The 'no conflicts' does not assure that the merge meets the intention of the developer. What is considered the base branch in this case?

I interpret 'jointly' to mean co-operation not finger-pointing.

quozl:
I don't understand why you want to know the base branch for merges, but it is *master*.

tony37:
Because master today may not be the same file as that used to construct 
the patch.

Tony

quozl:
Yes, of course, but it is *master* that the merge must be made to.  Nobody is interested in a patch to 0.108 to add a feature in the absence of all the subsequent patches that fix other problems.  Still, there are no *merge conflicts* as far as `git` is concerned; no other conflicting changes have been made to the code that this pull request touches.  Any problems you have with the four commits will be unrelated to a *merge conflict*.

tony37:
It is clear that no help with this is forthcoming. I'll try to see if 
this can be straightened out so that the PR can be merged (clearly no 
one believes the no conflict message and is willing to just 
automatically merge - me included).

Tony

quozl:
You are misinterpreting programmatic messages from GitHub, which for you may say one of these things;
- *This branch has no conflicts with the base branch*, or;
- *This branch has no conflicts with the base branch when rebasing*,

These messages are only indicative of the result of potential `git` commands, and have almost nothing to do with whether a pull request is ready to be merged by developers; except that if they are not present a developer will want to find out why.

What is needed now are changes to the pull request commits to;

- [ ] add a gsetting, with a default,
- [ ] use the gsetting,
- [ ] resolve reported incompatibility with TurtleBlocks, and;
- [ ] add evidence of testing with other activities.

iamutkarshtiwari:
Hi @quozl @tony37  Apologies for not being able to get involved in the discussion earlier. 

I tested this PR on XO 1.75 (Sugar 0.110.0.olpc.9) and I agree that it creates issues with TurtleBlocks in this listed Sugar version. It seems like that some changes made to the activity.py file after the PR are creating problems.

I think the best way to resolve this error would be to make these changes to the latest code of Sugar.

For gsettings, I can add that as well. Please let me know your suggestions

tony37:
Agreed.

I hope you have the chance to see how to retro-install to 0.110. I 
suppose a diff between the two versions of activity.py (ok 4) would 
reveal the issue.

Tony

iamutkarshtiwari:
However there is an exception to TurtleBlocks as it uses the activity.py from this https://github.com/sugarlabs/sugar-toolkit/pull/4 repository

iamutkarshtiwari:
@tony37 @quozl I just did some investigation and all the activities (both type of activities using activity.py from 'sugar-toolkit' & 'sugar-toolkit-gtk3') seem to work perfectly fine with this and https://github.com/sugarlabs/sugar-toolkit/pull/4 patch.

The problem (eg. crash of activities like TurtleBlocks) which I and Mr.Anderson are facing in XO with 0.110 installed is only the version sepecific. I'll provide with new patches for both 'sugar-toolkit' & 'sugar-toolkit-gtk3' by extracting activity.py for both module from 0.110 version.

There seems to be no issues in the latest versions of Sugar with these (https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/327 & https://github.com/sugarlabs/sugar-toolkit/pull/4) patches.

For ``GSettings``, I'll generate a new patch in a new branch and update the current PRs to use those gsettings values

Please Note: In latest ``Browse's`` activity code - https://github.com/sugarlabs/browse-activity/blob/master/webactivity.py#L349, there has been some changes made to override the jobject name on close using ``write_file()`` function which cause Browse to rename the jobject even after the user has provided a new name through the ``save as`` alert. I'll have generate a patch for Browse also so that it doesn't override the 'Save As' feature (when enabled).

iamutkarshtiwari:
@tony37 @quozl Here is the evidence to show that both of these patches work fine with gtk2 & gtk3 activities - https://www.youtube.com/watch?v=H_HjIi4XekQ

quozl:
@iamutkarshtiwari, thanks, looking forward to;

- [ ] commits for Browse, as a new pull request to https://github.com/sugarlabs/browse-activity,
- [ ] a pull request for https://github.com/sugarlabs/sugar to add the GSetting to the schema, with a default (disabled),
- [ ] new commits on this pull request #327 to use the GSetting to enable the feature,
- [ ] new commits on the other pull request https://github.com/sugarlabs/sugar-toolkit/pull/4 to use the GSetting to enable the feature.

tony37:
Thanks Utkarsh

It looks like you may have satisfied James Cameron. Well done!

Tony

icarito:
I'm glad a satisfactory merge is in sight.

Thanks for seeing it through till the end.

Sebastian

iamutkarshtiwari:
@quozl @tony37 Please review these-

1) GSettings added in sugar - https://github.com/sugarlabs/sugar/pull/750
2) Latest commits to use GSettings in 'sugar-toolkit-gtk3' - https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/327/commits/2caeead86aedb6efc9314232536165ab0b91b6af
3) PR to resolve ``Browse`` conflict with 'Save-As' - https://github.com/sugarlabs/browse-activity/pull/49

There is just one issue that I am facing for which I need your help. I am unable to figure out what kind of similar import is required for `from gi.repository import Gio` in **sugar-toolkit** as it is based on gtk2 imports like -

```
import gtk
import gobject
```

tony37:
Hi Utkarsh

Way out of my field. I suspect that like gobject, this was not used in gtk2.

Tony

quozl:
> I suspect that like gobject, this was not used in gtk2.

Wrong.  It is used.  `git grep gconf`.  `git grep gobject`.  Both show matches in sugar-toolkit.

quozl:
@iamutkarshtiwari, thanks, I've reviewed https://github.com/sugarlabs/sugar/pull/750 and https://github.com/sugarlabs/browse-activity/pull/49 but will wait for your https://github.com/sugarlabs/sugar-toolkit/pull/4 to check settings before reviewing the toolkit changes.

GConf is what you need; as "import gconf"; look at the existing code in https://github.com/sugarlabs/sugar-toolkit for references to the other settings near to the one you added in https://github.com/sugarlabs/sugar/pull/750/commits/eff97993e1af43350a6516b1fb501fcf82024438

Foreseeing the next review; you'll have to rebase this branch, reference the feature page, and please have a look at the variable names you've used; for a variable that stores a setting the name should be the same as the setting.

quozl:
Please also

- [ ] check the setting during write_file rather than save it at startup.

So that the setting can be exposed in the My Settings control panel and be immediately effective.

Also note the missing end of line at end of file on alerts.py, which GitHub shows as a red mark, and `flake8` reports as W292;

```
src/sugar3/activity/alerts.py:1:1: F401 '_' imported but unused
src/sugar3/activity/alerts.py:4:1: F401 'GObject' imported but unused
src/sugar3/activity/alerts.py:5:1: F401 'GLib' imported but unused
src/sugar3/activity/alerts.py:6:1: F401 'math' imported but unused
src/sugar3/activity/alerts.py:11:1: F401 'Icon' imported but unused
src/sugar3/activity/alerts.py:44:24: W292 no newline at end of file
```

The other `flake8` issues should also be worked.  I've not yet checked the other files; the red mark drew my attention first.

iamutkarshtiwari:
Here is the checklist - 
- [ ] Flake8 in alerts.py fixed
Please note - There are already existing flake8 errors before this PR. I would suggest fixing them in a separate PR rather than including in this one.
- [ ] Branch rebased to the latest commit.
- [ ] Variable name use for gsettings is `'_save_as_enabled'` since another variable with `_save_as` name has already been used in lot of places. And there are other existing gsettings variables in Sugar which go by this naming style - `gsetting_name_enabled` so should not be a problem.
-[ ] Commit message updated with reference page
- [ ] Check settings everytime rather than save it at startup

iamutkarshtiwari:
@quozl Please review the above checklist
Video for -
**GConf** for `sugar-toolkit` - https://www.youtube.com/watch?v=rgfw-MDL9LU
**GSettings** for `sugar-toolkit-gtk3` -https://www.youtube.com/watch?v=rNIauS31J5U

quozl:
Good progress, thanks.  I've done a visual review of the code changes;

In `__init__` the journal object is cloned if `save-as` is set; and later the alert handles the clone;
* what if `save-as` changes state between `Activity.__init__` and `Activity.close()`?  It doesn't look like the code handles that.  What do you think?

In `__save_response_cb` the alert is repeated if the title is empty;
* why not accept an empty title, as Sugar always did, and as my OLPC branch does, see https://bugs.sugarlabs.org/ticket/4940 "''My opinion is the bug seems like censorship; why can't the learner use a blank or empty name for a journal object? It is an unnecessary restriction on expression. Why can't Sugar just do what it is told? The learner can always discover how to rename again.''" ... also https://github.com/sugarlabs/sugar/pull/663 is still pending, which is one less patch for me to revert,
* as a side-effect of your blocking an empty title, clearing the title and then pressing Cancel brings the alert back again; the order of tests is wrong, please fix.

tony37:
There is no need to check the setting during a Sugar session. The setting was added to enable what I consider a legacy mode. Those who make the setting True should be presumed to know what they are doing.

tony37:
The title is shown in the Journal to identify an object for the user who wants to resume it. No title is not a sensible choice. This is like having an activity in the home view with no icon (and no tooltip?). A user can give a title with a single keystroke (title = 'a') so the requirement is hardly burdensome.

quozl:
Thanks for your opinion.  The setting may be changed during a session, as it will be available in the control panel.  An empty title is fine, it's how Sugar works; it's not a file name.

tony37:
If the title is blank, there will be no title displayed in the Journal 
view. I can't see any way in which an empty title is fine.

Tony

quozl:
It's how Sugar works, and I see no reason to change it.  As far as I can recall, an empty title has been valid and supported.  It is rare, because it takes care to achieve.

tony37:
That has happened to me when I wanted to change a title. I erased the 
title to write a new one and hit enter.  This can easily happen on an XO 
where the trackpad can cause strange things to happen. I would have 
appreciated a warning.

If the save as feature includes preventing a blank title being entered, 
consider that a part of the design. It does not stop the user going to 
the Journal and entering a blank title.  It does help the user to avoid 
entering a blank title by mistake.

Tony

quozl:
You are reinforcing my opinion; thank you.  The XO isn't the only hardware Sugar runs on.  An empty or blank title is easily recovered from by renaming the entry.  I don't think it is part of the design to prevent empty or blank title.  The patch doesn't even prevent a blank title, only an empty title.

tony37:
So you will allow #327 to protect the user from entering a  blank or 
empty title but will protect the inalienable right of a user to enter a 
blank or empty tiitle in the Journal.

Good decision.

Tony

quozl:
No, #327 is about forcing the learner to choose a name rather than accept the default.  If they want to choose a blank or empty name, that's fine.

tony37:
Total bs.

The purpose of #327 is to give the user an option to provide his own 
name. The default is the name of the activity which is meaningless. It 
should be something like 'untitled'.

I am sure you spend some time in the real world where a program asks you 
to supply a name for a file you are saving.

Try gimp. Open a new image. Save. Set the title from untitled.xcf to 
blank. Try to save.

Tony

quozl:
That's not how I see the use of this feature.  If you want to avoid empty or blank journal titles, that would be a different feature.  The world of professional programs used with filesystems instead of journals is not relevant.  The Sugar Journal is for learner reflection, and does not have the same constraints as a filesystem.  Empty or duplicate titles are fine.

tony37:
Again, your personal opinion. Again, 'that's not how I see the use of 
the feature'. What does it matter how you see the use of the feature?

Of course, programs like gimp are relevant. It is installed on most XOs 
with recent builds. Gnome makes the file system visible. Sugar commander 
and a number of activities expose the file system in Sugar.

How do see a learner reflecting on a Journal object with no title? Is 
this a zen thing where enlightenment comes from contemplating the 
meaning of 'None'.

All that is needed is for you to merge #327 and to be careful never to 
enable it on your own personal systems,

Tony

quozl:
Gimp is not relevant, because this patch is to Sugar, not Gnome.  You can't use Gimp with the Sugar Journal.  Sugar Commander is also not relevant.

Where you say "all that is needed is for you to merge", that's trivialising the developer collaboration and review process for Sugar.

You can always maintain your own fork if you like; and that is what I've already had to do because of my patches that Sugar Labs has not accepted, such as the performance improvements to the frame f6 key.  I respect that my patches may not have universal appeal; since they only helped XO users.

tony37:
As you may know, I submitted this feature for GSOC17.  Along with 
Sebastian Silva, I worked as a mentor on this project.

The feature was modeled on the way most document processors work. They 
offer the ability to save or save as the document. Save is used when the 
document has already been saved and so uses the same path. Save as gives 
the user the chance to change the name and/or path. Normally a save 
issued without a prior save as is treated as save as. AFAIK, this 
feature did not appear in Sugar because of the desire in the HIG to 
avoid the file menu and the appearance of a 'desktop'. It is my 
understanding that search tags (a la Picasa) would offer the 
organization normally found in directories. However, this was not 
implemented.

Gimp is an example of the behavior expected of the feature. Sugar 
commander is only relevant to your comment that Sugar does not have a 
file system. Of course it does - it choses not to make that visible to 
its users.

I completely agree with James Simmons. It should have been. We want our 
users to grow into a confident understanding of how computers work. This 
includes an understanding of the file system. I would like to see the 
Documents and removable devices shown in the Journal as a hierarchy. 
This would not be difficult technically and would facilitate dealing 
with usb drives with many files. One technique could be to use the 
directory names on a usb stick as tags The user could search for files 
in a directory using the name as a tag.

I would certainly appreciate improvements to performance of the frame 
key. In writing code in Sugar, it is essential to open and close it 
frequently to switch between activities.  You should certainly raise the 
issue again so we have a chance to benefit. Only applying to the XO 
covers the overwhelming majority of our users. So I really don't 
undertand why you met any resistance.

I would certainly like to understand this developer collaboration and 
review process. Utkarsh has jumped through every hoop. The PR shows that 
all issues raised have been satisfied. To accommodate those who would 
prefer not to see this change in the Sugar UI, it is controlled by a 
setting with the default that it is not enabled.

The PR works much the same way as the corresponding feature for 
screenshots. I don't even know if Utkarsh requires a non-blank 
screenshot title. The setting of the default to false was used with the 
PR for the 'resume' feature which has been merged.

I assume this process is intended to ensure that a PR request does not 
break Sugar, is compatible with Sugar and its HIG, and meets reasonable 
and publicly documented technical standards.

PR #327 meets these criteria. The only roadblock to its merger seems to 
be you and your personal opinions.

Tony

icarito:
El 26/04/17 a las 01:45, tony37 escribió:
> Total bs.
Tony, if we had a proper code of conduct, your way of expressing
yourself shouldn't be accepted behaviour, in my humble opinion. Please
note that in writing across cultural differences, your tone is lost and
so you come across as rude, at the least.

> I would certainly like to understand this developer collaboration and
review process.

Wouldn't we all? We have a shortage of commit-ready developers, so let's
treat them well.

> PR #327 meets these criteria. The only roadblock to its merger seems to
be you and your personal opinions.

From the Sugar Labs perspective, he is a volunteer and is entitled to
his opinions. To merge requires active action from a developer who
cares. Not simply "removing roadblocks". There is nobody blocking you or
me from forking and merging anything we like.

Sebastian

iamutkarshtiwari:
> Good progress, thanks. I've done a visual review of the code changes;
> 
> In __init__ the journal object is cloned if save-as is set; and later the alert handles the clone;
> 
> what if save-as changes state between Activity.__init__ and Activity.close()? It doesn't look like the code handles that. What do you think?

Nice catch! I'll provide an updated commit for this fix.

> In __save_response_cb the alert is repeated if the title is empty;
> 
Initially it was required as we are blocking the user from providing any blank title.
> why not accept an empty title, as Sugar always did, and as my OLPC branch does, see https://bugs.sugarlabs.org/ticket/4940 "My opinion is the bug seems like censorship; why can't the learner use a blank or empty name for a journal object? It is an unnecessary restriction on expression. Why can't Sugar just do what it is told? The learner can always discover how to rename again." ... also sugarlabs/sugar#663 is still pending, which is one less patch for me to revert,
> as a side-effect of your blocking an empty title, clearing the title and then pressing Cancel brings the alert back again; the order of tests is wrong, please fix.

I feel that if the community agrees to allowing user to provide blank entries to journal objects then I can modify the patch. And going by @davelab6 comment https://github.com/sugarlabs/sugar/pull/663#issuecomment-205327648, he seems fine with this. Let's see what others( @walterbender @i5o @samdroid-apps @icarito)  think about it.

@tony37 We are almost at the verge of completion and after gaining the community consensus on this 'blank title' idea, we can get it merged soon :)

quozl:
@iamutkarshtiwari, thanks, looking forward to updated commit.  Let me know if you run out of time and I can look into it; don't want to undermine your work, but I don't want to lose it either.

For blank or empty titles, let's concentrate the discussion at https://github.com/sugarlabs/sugar/pull/752

quozl:
Tested on Ubuntu 16.04 with b8cf4a26a248e4e82d4a9c45001d66a06379db27 and https://github.com/sugarlabs/sugar/pull/750/commits/3df38ff1a5e626c8a0db71357b9c597c507000cc with GTK+ 3 activities.

**appearance issues**

- [ ] the alert says "Provide a _name_ for the _project_", but it should be a name for the journal entry; because of (a) consistency with the [Help Activity for Journal](https://github.com/godiard/help-activity/blob/master/source/journal.rst), and (b) the word _project_ now has a different meaning after the project support in the journal landed for 0.110,

- [ ] the second button says "Quit", but it does not say that any cloned journal entry will be deleted; perhaps it should say something else?

**interaction issues**

- [ ] when an activity instance is created (i.e. start new, no prior journal entry), and then Stop is pressed, followed by Quit, a journal entry is left behind; this does not comply with the feature page which says 

> for a new document (start new), when the user quits the activity, a popup provides an opportunity for the user to give the document a name. The _user can also decide not to save it_.

- [ ] when an journal entry is resumed, and then Stop is pressed, followed by Save, without changing the name, a new journal entry is created; this does not comply with the feature page which says

> for an existing document, when the user quits the activity, a popup provides an opportunity to give it a new name. In this case, it will be saved under the new name and the original will be preserved. If the _user chooses not to provide a new name, the document is saved replacing the original_.

- [ ] when the name is set to an empty string, and Save is pressed, the alert is re-issued without explanation or warning,

- [ ] when the name is set to an empty string, and Quit is pressed, the alert is re-issued; it should Quit instead, as mentioned earlier in the week,

- [ ] the name can be set to a blank space, but not an empty string; while we are still debating the other issue, please at least be consistent within your commit, :grin:,

- [ ] when an activity Stop button is pressed, a busy cursor is shown during the Save As alert, and if Escape is pressed the cursor stays busy; the busy cursor should be turned off during the alert, or turned on after the alert is dismissed.

- [ ] the `Enter` key was not accepted; the key should do the same as the Save button,

**configuration issues**

- [ ] setting `save-as` should be in `org.sugarlabs.journal` as it isn't an attribute of the user `org.sugarlabs.user`,

- [ ] same can be said for `resume` from 695889addd53c1b197b35a4828b56d9003fdb720 which was a slip-up on our part,

**good results**

For your interest, things that worked fine;
- the `Esc` key to dismiss the alert and cancel the Stop,
- logout was not delayed,
- an instance of Write was kept in the Journal despite logout,
- changing the cloned entry name using the Journal f5 key,

Also intended design that might need to be documented or clarified; when an activity is used to resume a journal entry, then the Journal is displayed with f5, there are now two journal entries with the same name, the latest is the clone, but it has the same name as the older entry.

@iamutkarshtiwari, you've done well.  Please ask any questions if you need more detailed explanation.

tony37:
Thanks for your careful test.

Perhaps 'Give a name to your work'. What Quit means is the Journal 
remains unchanged. Since Sugar now saves an object early and often, 
restoring the Journal to its pre-activity state will require an explicit 
delete.

As I understand your proposal, a valid title must not consist only of 
whitespace. A test 'if title.strip() should work.

As easy route would be to treat entry of an invalid title the same as 
'Quit'. However, this option clearly risks the user losing his work if 
the save key was hit inadvertently before entry of the new title was 
completed. This risk goes up if enter is treated as save.

On entry of an invalid title, perhaps the alert could change wording to 
something like; 'Your work does not have a name. Please give a name to 
your work.

I agree about 'enter'. In workshops I make a big point that the computer 
does not carry out a command until you press enter.

I don't understand your comment about org.sugarlabs,journal. The feature 
is implemented as a change to activity.py and has nothing to do with the 
Journal View.

The two names represent the fact that Sugar now saves early and often. 
It also highlights the flaw in Sugar which over-writes the resumed object.

Tony

tony37:
My mistake. I had forgotten that org.sugarlabs.xxxx is how the settings are organized. Certainly putting the setting in org.sugarlabs.journal is no problem.

I also forgot about the 'statistics' requirement. The Journal object metadata should not be deleted because that is a record of the way the user employed Sugar. To minimize the size of the datastore, the file should not be saved.

iamutkarshtiwari:
@quozl I am trying to capture keyboard focus (get cursor to be active inside) inside Gtk.Entry() text area so that I can detect 'Enter' keypress. I tried this -

```
entry = Gtk.Entry()
entry.set_can_focus(True)
...
...
entry.show_all()
entry.grab_focus()
```
But I am still unable to grab keyboard focus inside Gtk.Entry. Could you please help?

iamutkarshtiwari:
> the second button says "Quit", but it does not say that any cloned journal entry will be deleted; perhaps it should say something else?

I am thinking of something like this - 
![screen shot 2017-04-29 at 8 44 11 pm]
screen shot 2017-04-29 at 8 52 26 pm

iamutkarshtiwari:
@quozl I am not squashing the commits now, they will help you track my frequent changes. I'll squash all the commits in the end.

iamutkarshtiwari:
> when an activity instance is created (i.e. start new, no prior journal entry), and then Stop is pressed, followed by Quit, a journal entry is left behind; this does not comply with the feature page which says
> for a new document (start new), when the user quits the activity, a popup provides an opportunity for the user to give the document a name. The user can also decide not to save it.
>  when an journal entry is resumed, and then Stop is pressed, followed by Save, without changing the name, a new journal entry is created; this does not comply with the feature page which says

Descript updated - https://wiki.sugarlabs.org/go/Features/Save-As#Detailed_Description

iamutkarshtiwari:
> when an activity Stop button is pressed, a busy cursor is shown during the Save As alert, and if Escape is pressed the cursor stays busy; the busy cursor should be turned off during the alert, or turned on after the alert is dismissed.

@quozl  Could you please suggest an starting point for this fix? This effect seem to trigger from some other place not activity.py

quozl:
Thanks.

> still unable to grab keyboard focus inside Gtk.Entry. Could you please help?

Python GI API Reference documentation for `Gtk.Widget.grab_focus` says in part;

> The widget also needs to be realized and mapped. This is indicated by the related signals. Grabbing the focus immediately after creating the widget will likely fail 

But as far as I can see, you have only called `show` for the widget, you haven't connected to the `realize` or `map` signals.  The widget won't be realized and mapped for some time after return from function.

> > busy cursor is shown
>
> please suggest an starting point for this fix? This effect seem to trigger from some other place not activity.py

It is done by the call to `set_cursor`, see [activity.py close](https://github.com/iamutkarshtiwari/sugar-toolkit-gtk3/blob/feature-saveAsPopup/src/sugar3/activity/activity.py#L1204).

Re: https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/327/commits/68f0bde7fcdfeea3ae8bab038b921f2f73565341

- [ ] in one part you say "Cancel", but in another "Don't Save"; you may solve by removing the "(Press \'Esc\' to Cancel)" ... we do not usually describe the keyboard shortcuts in the UX.

Re: https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/327/commits/e5b9c4b5c15d13a14a021ac45ae647040d21bce3

- [ ] the check for Return is not protected by a check for `self._save_alert`, and so it will prevent using Return key in any activity?

icarito:
@iamutkarshtiwari : 
>  I am trying to capture keyboard focus (get cursor to be active inside) inside Gtk.Entry() text area so that I can detect 'Enter' keypress.

Try putting the call to get_focus inside a [`GObject.idle_add`](http://www.programcreek.com/python/example/79693/gi.repository.GObject.idle_add) call.

icarito:
Another reference: https://dev.gajim.org/gajim/gajim/commit/3fd6afb3112786a4235b52d5c5805d842325b5ed

iamutkarshtiwari:
Wow, it works! Thank you @icarito :)

iamutkarshtiwari:
@quozl I'll update https://github.com/sugarlabs/sugar-toolkit/pull/4 accordingly after this PR is completed

quozl:
> putting the call to get_focus inside a `GObject.idle_add` call.

No, that's wrong, and may create a race condition.  It may work most of the time, but sometimes it will fail, if the graphics device is momentarily delayed, and especially on multicore systems where the GTK+ program may move to idle processing before another core has finished graphics device update.  It is better to connect to the `realize` or `map` signal, and that's what the documentation says to do.

Also, your 91d2f74 is ambiguous; the idle function is called repeatedly until it returns False, and `grab_focus` is not specified to return False.  That it does seem to is somewhat accidental.

quozl:
Also in 97c783614e81f7bc75e86fac4248f4cbba4f81ff when `save-as` is enabled,
- there is no busy cursor during close; this is a regression, which you can fix by making the cursor busy again when the alert button 'Save' is pressed,
- the `_closing` signal is emitted early, which causes a `save_title` in widgets.py to run, but is that necessary?
- the activity `can_close` is called only once, yet the learner can continue to interact with the activity (which I've just tested again) and may invalidate `can_close`; this is a regression on the meaning of `can_close`, which you can fix by either making the alert modal (which would be unfortunate), or calling `can_close` again sometime later.

For your testing, among the activities that implement `can_close` are Browse, Chat, Develop, FractionBounce, GetBooks, ImageViewer, Jukebox, Maze, Physics, Read, and TurtleBlocks.

I'm having trouble following the new logic path in `close`; it seem convoluted to call `_alert_confirmation` only for that method to find out that the title has been changed, then set a flag `_pre_naming` which is tested again later in `close`.

tony37:
Thanks for this testing. You are showing some serious ambiguity in the 
core implementation of activity.py

A return of True from can_close should end processing except write_file.

Aside from user 'ctrl' + q or quit, the activity could be closed as a 
result of the end of the Sugar session by shutdown or logout or by 
shutdown of the system because of lower battery power.

As I recall, activity developers should assume the object changes made 
in write_file will be saved.

For the quiz activity, I have had to set a flag in can_close because the 
information needed by write_file is not available until the quiz 
responses are complete.
This no-ops the write_file calls until after can_close has been called.

Tony

iamutkarshtiwari:
> Also in 97c7836 when save-as is enabled,
> 
> there is no busy cursor during close; this is a regression, which you can fix by making the cursor busy again when the alert button 'Save' is pressed,

Fixed
> the _closing signal is emitted early, which causes a save_title in widgets.py to run, but is that necessary?
_closing signal is now emitted only when the user takes an actions (Save/Don't save) from alert
> the activity can_close is called only once, yet the learner can continue to interact with the activity (which I've just tested again) and may invalidate can_close; this is a regression on the meaning of can_close, which you can fix by either making the alert modal (which would be unfortunate), or calling can_close again sometime later.

After checking I found out that the 'can_close()' is called `everytime` 'close()' functions executes i.e. everytime 'Stop' button is pressed. 

> For your testing, among the activities that implement can_close are Browse, Chat, Develop, FractionBounce, GetBooks, ImageViewer, Jukebox, Maze, Physics, Read, and TurtleBlocks.
> 
> I'm having trouble following the new logic path in close; it seem convoluted to call _alert_confirmation only for that method to find out that the title has been changed, then set a flag _pre_naming which is tested again later in close.

_alert_confirmation code moved inside 'close()' to remove ambiguity

quozl:
@iamutkarshtiwari, thanks.  I've tested all day and found many problems.  It was easier to fix the problems and rework the code than go into detail explaining what I've fixed.  I've pushed two branches:

* https://github.com/quozl/sugar-toolkit-gtk3/commits/2017-123-save-as-refactoring
* https://github.com/quozl/sugar-toolkit-gtk3/commits/2017-123-save-as-refactoring-rebased

Please consider;
* do a code review of the overall difference from master at cddaf790eddf3e1077b6b646bf61c0faf6ebf785,
* fetch, checkout and test either of my branches,
* if you agree with them, or they aren't too far from what you want, cherry-pick into your branch my four patches on top of yours, 751c178d23cffb3fed016ef5e0dfb399a2b2d542, 011ccff02b3dd805b24d1f9a418d7f31a9e93c37, 36b7f45726741c9c41fce1703d5a89fa209e4148 and 2cb0b33e77ad05b60290f099c820b7c76d1d36f6,
* look at the FIXME markers I've added; in particular the Write activity doesn't seem to work with the focus grab, and there are some minor private member accesses in activity/alerts.py.

quozl:
Thanks for taking my changes, you've encouraged me to look further at the problems.

I've pushed 11 patches to my branch, please review and cherry-pick them;
* https://github.com/quozl/sugar-toolkit-gtk3/commits/2017-123-save-as-refactoring

Progress was;
* use `datastore.copy()` so that the clone will have all the statistical metadata of the original object, otherwise metadata was being lost,
* add context sensitive tooltips and button labels, because the meaning of the buttons changes slightly depending on the context,
* fix several problems when used with shared activities, like resuming a shared activity didn't make it shared again, and every participant underwent the stop alert even if the name had been changed,
* fixed the minor private member accesses in `alerts.py` by moving code into `alert.py`; alerts now support a Gtk.Entry,

Also tested with shared activities; in particular Maze and Chat.

quozl:
@iamutkarshtiwari, thanks for taking my changes; have you reproduced the problem with focus on the Write activity with these patches applied?  Any ideas on what causes it?  I've not seen any other activity affected.  I'm inclined to let it pass.

quozl:
Rebased and merged as 926a2623067cdbeadeec4b6f8d447874f34a02d3, 83697d21f914613d361350fa1c63cd44c1d1a0f5, 286387b22e70ffe15f51e782ebf48a68ac3646f1, b1d11297593c57a0729f551b2c5f9d716aa24710, 6bcd6642896e0f75922ac25c367e5e128b67bd01, 3a574ae77888b77003f567383eec3636e9b25945, and ba6993d70522524941fec6721d38ff6d79ec4a68.

tony37:
Thanks for your effort and patience in making this possible. I think you 
helped this feature become more useful and robust.

Tony

iamutkarshtiwari:
@tony37 @quozl Thank you for your support and guidance. I am happy that it got  merged! :)

PR #331, opened 2016-07-14
Bundlebuilder patches, including generating AppStream data files!

samdroid-apps:
Probably replaces https://github.com/hughsie/appstream-glib/pull/111

### DIFF
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -461,7 +461,7 @@ class Installer(Packager):
             ET.SubElement(root, 'url', type='homepage').text = \
                 info.get('Activity', 'url')
         if info.has_option('Activity', 'repository_url'):
-            ET.SubElement(root, 'url', type='repository').text = \
+            ET.SubElement(root, 'url', type='bugtracker').text = \
                 info.get('Activity', 'repository_url')
 
         path = os.path.join(prefix, 'share', 'metainfo',
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -1,4 +1,5 @@
 # Copyright (C) 2008 Red Hat, Inc.
+# Copyright (C) 2016 Sam Parkinson <sam@sam.today>
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


davelab6:
Nice!

src/sugar3/activity/bundlebuilder.py
```diff
@@ -345,6 +411,61 @@ def _install_desktop_file(self, prefix, activity_path):
...
+            if not info.has_option('Activity', name):
+                print('[WARNING] Activity needs more metadata for AppStream '
+                      'file')
+                print('  Without an AppStream file, the activity will NOT '
+                      'show in software stores!')
+                print('  Please `pydoc sugar3.activity.bundlebuilder` for'
+                      'more info')
+                return
+
+        # See https://people.freedesktop.org/~hughsient/appdata/ for spec
```
ximion: Please don't use the old location, and rather the "new" spec:
https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html, which documents all valid tags excessively, or the quickstart guide: https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps

src/sugar3/activity/bundlebuilder.py
```diff
@@ -345,6 +411,61 @@ def _install_desktop_file(self, prefix, activity_path):
...
+                ET.SubElement(e, 'image').text = screenshot
+
+        if info.has_option('Activity', 'url'):
+            ET.SubElement(root, 'url', type='homepage').text = \
+                info.get('Activity', 'url')
+        if info.has_option('Activity', 'repository_url'):
+            ET.SubElement(root, 'url', type='repository').text = \
+                info.get('Activity', 'repository_url')
+
+        path = os.path.join(prefix, 'share', 'appdata',
```
ximion: Please don't use the legacy path if possible (should be "/usr/share/metainfo/") - use the legacy path only if you need backwards compatibility with old AppStream tooling.

samdroid-apps:
Thanks for the review @ximion!

src/sugar3/activity/bundlebuilder.py
```diff
@@ -345,6 +411,61 @@ def _install_desktop_file(self, prefix, activity_path):
...
+                e = ET.SubElement(ss_root, 'screenshot')
+                if i == 0:
+                    e.set('type', 'default')
+                ET.SubElement(e, 'image').text = screenshot
+
+        if info.has_option('Activity', 'url'):
+            ET.SubElement(root, 'url', type='homepage').text = \
+                info.get('Activity', 'url')
+        if info.has_option('Activity', 'repository_url'):
+            ET.SubElement(root, 'url', type='repository').text = \
```
ximion: A repository URL type does not exist...
But you might want to add one for the bugtracker.
See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url for details on `<url/>`

PR #338, opened 2016-08-14
Proper deprecation of Palette.popdown state argument

samdroid-apps:
The popdown state argument was removed in 90cec5b, as the palette
popdowns were unified.  However, this is a public api.  Removing
the argument caused exceptions to be raised, breaking applications.
Most uses of the state argument were for the SECONDARY state, which
is now the default behaviour.

This commit re-adds the argument with a deprecation notice.

### DIFF
--- a/src/sugar3/graphics/palette.py
+++ b/src/sugar3/graphics/palette.py
@@ -223,7 +223,19 @@ class Palette(PaletteWindow):
 
         PaletteWindow.popup(self, immediate)
 
-    def popdown(self, immediate=False):
+    def popdown(self, immediate=False, state=None):
+        '''
+        Popdown (or show the full contents of) the palette.
+
+        Keyword Args:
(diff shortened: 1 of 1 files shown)


src/sugar3/graphics/palette.py
```diff
@@ -223,7 +223,19 @@ def popup(self, immediate=False):
...
+    def popdown(self, immediate=False, state=None):
+        '''
+        Popdown (or show the full contents of) the palette.
+
+        Keyword Args:
+            immediate (bool): if True, the palette will be shown instantly (as
+                if the user right clicked the item).  If False, the palette
+                will be shown after the usual activation wait time.
+
+        .. deprecated:: 0.109.3
```
quozl: There's no 0.109.3 in your [release roadmap](https://wiki.sugarlabs.org/go/0.110/Roadmap), perhaps you mean 0.109.2.0 feature freeze on 15th, or just use 0.109?

PR #340, opened 2016-10-08
sugar-activity: typo

quozl:
(Quite an old change found in OLPC branch but not yet upstream.)


leonardcj:
Not a translatable string as I see it.

quozl:
Yes, that's true.

leonardcj:
In that case, no objection :-)

quozl:
Heh. :grinning:

PR #343, opened 2016-11-09
Fix screen resize race

quozl:
When an external monitor is connected or disconnected, about 25% of the time an activity does not resize correctly; a `size-changed` signal occurs but the workarea geometry has not changed.

Underlying problem is an update race in `get_monitor_workarea`, so the fix is to call `get_monitor_geometry` for obtaining the width and height.

Can be reproduced easily without an external monitor by switching display panel resolution repeatedly;
```
    xrandr --output eDP1 --mode 1024x768
    xrandr --output eDP1 --mode 1366x768
```
Part of a fix for https://bugs.sugarlabs.org/ticket/4968

----
Related pull request https://github.com/sugarlabs/sugar/pull/732 can be merged in any order with this pull request; the patches are independent but solve the same problem in two different contexts.

### DIFF
--- a/src/sugar3/activity/activity.py
+++ b/src/sugar3/activity/activity.py
@@ -626,12 +626,12 @@ class Activity(Window, Gtk.Container):
 
     def _adapt_window_to_screen(self):
         screen = Gdk.Screen.get_default()
-        workarea = screen.get_monitor_workarea(screen.get_number())
+        rect = screen.get_monitor_geometry(screen.get_number())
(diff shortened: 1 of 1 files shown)


quozl:
Rebased to master.

i5o:
Merged
https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/a6e62a52b6eae4130caa0aa09f87def3ab18975f

PR #355, opened 2017-01-02
Write documentation for sugar3.power

quozl:
- get it done before someone else tries it,
- tested using `make-doc.sh`,
- following [documentation style](https://wiki.sugarlabs.org/go/Features/Sugar3_Docs/How_To_Write); and,
- with method references [Cross-referencing Python objects](http://www.sphinx-doc.org/en/1.5.1/domains.html#cross-referencing-python-objects).

### DIFF
--- a/src/sugar3/power.py
+++ b/src/sugar3/power.py
@@ -15,6 +15,17 @@
 # Free Software Foundation, Inc., 59 Temple Place - Suite 330,
 # Boston, MA 02111-1307, USA.
 
+"""
+The power module provides an interface to *powerd*, a daemon that
+manages the aggressive suspend and wakeup policies for early OLPC
+laptops.
+
+The module does nothing if *powerd* is not present.  *powerd* is not
+required on laptops other than OLPC XO-1, XO-1.5, XO-1.75 and XO-4.
+Distributions of Sugar on other hardware need not include the *powerd*
+package.
+"""
+
 import os
 import logging
 
@@ -24,6 +35,13 @@ _power_manager = None
 
 
 def get_power_manager():
+    """
+    Get the power manager instance.  Only one instance exists, and
+    will always be returned.
+
+    Returns:
+        an instance of :class:`sugar3.power.PowerManager`.
+    """
     global _power_manager
     if _power_manager is None:
         _power_manager = PowerManager()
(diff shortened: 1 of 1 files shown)


src/sugar3/power.py
```diff
@@ -15,6 +15,17 @@
...
 
+"""The power module provides an interface to *powerd*, a daemon that
+manages the aggressive suspend and wakeup policies for early OLPC
+laptops.
+
+The module does nothing if *powerd* is not present.  *powerd* is not
+required on laptops other than OLPC XO-1, XO-1.5, XO-1.75 and XO-4.
+Distributions of Sugar on other hardware need not include the *powerd*
+package.
+
```
i5o: Can we remove this newline?

src/sugar3/power.py
```diff
@@ -15,6 +15,17 @@
 # Free Software Foundation, Inc., 59 Temple Place - Suite 330,
 # Boston, MA 02111-1307, USA.
 
+"""The power module provides an interface to *powerd*, a daemon that
```
i5o: As I've seen, most of new documentation is written after the 
`"""`
can you change it?

src/sugar3/power.py
```diff
@@ -24,16 +35,48 @@
...
+    continue to dim and blank the display.
+
+    Both the :py:meth:`inhibit_suspend` and :py:meth:`restore_suspend`
+    methods are reference counted; automatic idle suspend is not
+    restored until the same number of calls to restore are made.
+
+    *powerd* is resilient against failure to restore automatic idle
+    suspend; it verifies an inhibit request and deletes it if the
+    requesting process has terminated.
+
```
i5o: same with this newline and the others

quozl:
Sure, easily done, 3d4bf93.  The format came from emacs fill paragraph feature in python mode and I didn't have any reason to object to it.  But "most new documentation" is a good reason.

quozl:
Rebased to master.

src/sugar3/power.py
```diff
@@ -47,9 +91,18 @@ def __del__(self):
...
+        toolkit?  Yes.  For future use by a toolkit with more
+        resilient collaboration.
+
+        Returns:
+            True
+        """
         return True
 
     def inhibit_suspend(self):
+        """Inhibit automatic idle suspend until restored."""
```
i5o: We also leave a newline here; as seen in
https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/src/sugar3/graphics/xocolor.py#L279
https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/src/sugar3/graphics/toolbutton.py#L206
etc

quozl:
Fixed in 3659570.  A pity, as vertical space is always the limiting factor when a new developer is trying to comprehend code they've never seen before.

i5o:
Merged 
https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/99130b02639600f2645667eb514ef0eea4df0fa5

PR #359, opened 2017-01-07
Added example and documentation for colorbutton

Hrishi1999:
The example is tested and it works, though I have a doubt. If I test this example in PEP8, it will give warning about there is no last line at the end of file but if I use flake8, it gives me warning about and empty last line at end of file. 
Please review it.
Update : I have squashed multiple commits.

### DIFF
--- a/examples/colorbutton.py
+++ b/examples/colorbutton.py
@@ -1,0 +1,35 @@
+from gi.repository import Gtk
+
+from sugar3.graphics.toolbarbox import ToolbarBox
+from sugar3.graphics.colorbutton import ColorToolButton
+
+
+import common
+
+
+test = common.Test()
+test.show()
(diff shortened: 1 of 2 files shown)


i5o:
Please notify the user (maybe by printing it) when the color is changed http://people.sugarlabs.org/ignacio/pi/iaWE678siU.txt

I will wait for @quozl review, as usual he knows this stuff :)

Hrishi1999:
Thanks for the snippet (print color_button...and not button). Also, i would go with print (earlier I was adding an alert), but now I would go with just print.
Will submit changes soon.

i5o:
Thanks @Hrishi1999 looks good to me.

Can you squash them and make a single description? Thanks again

PR #360, opened 2017-01-11
Example fixes

i5o:
- replaced Gtk.HBox/Gtk.VBox by GtkBox as seen in [1] and [2]
GtkHbox and GtkVBox are deprecated
- use delete-event instead of destroy event in GtkWindow
- replaced icon_size for pixel_size where used
it's also deprecated as seen in commit [3]
- use sugar_theme in all examples; set it manually not by just
importing common module
- fixed GtkBox.pack_start arguments
- flake/pep8 fixes

[1] https://developer.gnome.org/gtk3/stable/GtkHBox.html
[2] https://developer.gnome.org/gtk3/stable/GtkVBox.html
[3] 5802d67ee107ef82963770257071a779c4741a21

### DIFF
--- a/examples/alert.py
+++ b/examples/alert.py
@@ -1,10 +1,7 @@
 from gi.repository import Gtk
-
 from sugar3.graphics.alert import TimeoutAlert
-
-
-def _destroy_cb(widget, data=None):
-    Gtk.main_quit()
+from common import set_theme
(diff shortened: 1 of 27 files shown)


quozl:
Reviewed a8b0ee7.

PR #361, opened 2017-01-31
Added the missing get_locale_path() function from gtk2 version

yatinmaan:
The sugar3.activity.i18n module is missing the get_locale_path() function present in sugar.activity.i18n.

### DIFF
--- a/src/sugar3/activity/i18n.py
+++ b/src/sugar3/activity/i18n.py
@@ -18,7 +18,11 @@
 # Boston, MA 02111-1307, USA.
 
 from gettext import gettext
+import locale
+import os
 import struct
+import sys
+
 
 import dateutil.parser
 import time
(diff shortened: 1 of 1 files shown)


quozl:
Thanks, that's interesting, but I don't think it is needed.  Why do you think it is needed?  What activities have you tested with this change?  We have several ways in which internationalisation is implemented in activities, and looking through the history the early import of locale might break some activities or activity start methods.

See for example this discussions and later commit;
 * http://lists.sugarlabs.org/archive/sugar-devel/2012-May/037678.html
 * 6330204e919a047c0b0f3579da36197630e8e400

Please also show activities that would require this; and those that don't require it; so we can work together to change the activity code.

If it is an activity you are porting to Gtk3, then look at other Gtk3 ports to see how i18n is done?

Also, by making your ported activity require the _next_ version of sugar-toolkit-gtk3 you will significantly restrict user base.  Activities should be as portable as possible across toolkit versions.

yatinmaan:
I asked about this in IRC and was told that this function was probably overlooked during porting so decided to add it. I didn't know that the removal was intentional. I would have opened an issue but Issues are not enabled on this repo.

Interestingly, I wanted to get the locale path while working on the "untranslated languages in settings menu" bug but was unaware of this function's existence then and came up with my own hacky solution of using an gettext internal function 
```python
locale=gettext._localedirs[gettext.textdomain()]
```
Maybe we could document this somewhere as I had a hard time finding it.

PR #362, opened 2017-02-12
Allow to disable installing .desktop files

City-busz:
Similar to custom mime types, allow to disable it.


tony37:
What is the impact of this on Sebastian Silva's technique to enable 
activities through the gnome desktop?

Tony

PR #363, opened 2017-03-27
Prepend instead of append to path

quozl:
When an activity is launched, an import should first search the activity bundle source rather than `/usr/bin`.

Fixes the need to change activity sources every time a system package conflicts with an import.

May impact any activity that relies on a system package being in search path before the bundle source.  Forty activities were tested, and none relied on this.

See also https://github.com/sugarlabs/Pippy/pull/45

CC @icarito.


icarito:
It's awesome that you discovered the culprit! I had had issues with other activities as well.
Reviewed.

PR #364, opened 2017-04-15
LINGUAS update, Pootle push

leonardcj:


quozl:
Nak, breaks build.

`LINGUAS` contains `su` and `tyv` but there are no corresponding files in `po/` with these names, and as a result the `make distcheck` step fails.  Perhaps you missed `git add`?

Workaround is to remove `su` and `tyv` from `LINGUAS`.

leonardcj:
Dropped su and tyv

quozl:
Thanks.  Reviewed.  Doesn't break build now.

PR #365, opened 2017-05-06
Pootle push

leonardcj:


quozl:
Thanks, that's good.  Reviewed.  Will merge.  Process feedback for future commits;

- use the word _translation_ in addition to, or rather than, _pootle_, so that our new developers doing git bisects don't have to try to find what the word _pootle_ means,

- use a branch other than _master_, as described in our [guide to contributing](https://developer.sugarlabs.org/contributing.md.html); mainly so that your practice does not encourage our new developers to use _master_, but also because GitHub will automatically offer to create a pull request from a branch and not from _master_, and so that your clone can support multiple simultaneous pull requests should you ever need to do that.

Hrishi1999:
I will be doing more Gujarati translations later.  I had time to only do this much :(

quozl:
@hrishi1999, thanks for the translation.  I didn't know you were involved, the Last-Translator metadata had not been changed for `po/gu/po`; or I'm looking at the wrong file.  We all appreciate translations, do continue to find time if you can.

PR #379, opened 2018-02-07
Fix: #366 sugar3.graphics.alert.TimeoutAlert does not expose timeout condition 

### LINKED ISSUE #366: sugar3.graphics.alert.TimeoutAlert does not expose timeout condition
Documentation for TimeoutAlert says the response ID should be -1 on timeout, but the response ID is OK instead.  Thie goes back to the GTK+ 2 toolkit code as well.

Activities may have been coded to rely on this.

Either;
* fix the documentation, or
* deprecate the API and make a new one, or
* fix the code to return a response ID of -1 on timeout, and fix any activities that relied on OK.

Pro-Panda:
**Effect of this PR:**
 - `_TimeoutAlert` returns a response ID of -1 on timeout
 - Checked for activities using this response via [this](https://github.com/search?q=org%3Asugarlabs+TimeoutAlert&type=Code). Turns out none of the activities use the response emitted on timeout.

@quozl Kindly review

### DIFF
--- a/examples/alert.py
+++ b/examples/alert.py
@@ -1,25 +1,33 @@
 from gi.repository import Gtk
 from sugar3.graphics.alert import TimeoutAlert
 from common import set_theme
+
+
 set_theme()
-
-
-def __start_response_cb(widget, data=None):
-    print 'Response: start download'
-
-
 w = Gtk.Window()
 w.connect("delete-event", Gtk.main_quit)
-
 box = Gtk.Box(orientation=Gtk.Orientation.VERTICAL)
 w.add(box)
 
-alert = TimeoutAlert(9)
-alert.props.title = 'Download started'
-alert.props.msg = 'Sugar'
+# Create a TimeoutAlert Object
+alert = TimeoutAlert(5)
+alert.props.title = '<Alert Title>'
+alert.props.msg = '<Alert Message>'
+
+# Add Timeout Object to the box
 box.pack_start(alert, False, False, 0)
-alert.connect('response', __start_response_cb)
 
+
+# Called when an alert object throws a response event.
+def __alert_response_cb(alert, response_id):
+    if response_id is Gtk.ResponseType.OK:
+        print 'Continue Button was clicked.'
+    elif response_id is Gtk.ResponseType.CANCEL:
+        print 'Cancel Button was clicked.'
+    elif response_id == -1:
+        print 'Timeout occurred'
+
+
+alert.connect('response', __alert_response_cb)
 w.show_all()
-
 Gtk.main()
--- a/src/sugar3/graphics/alert.py
+++ b/src/sugar3/graphics/alert.py
@@ -406,7 +406,7 @@ class _TimeoutAlert(Alert):
         self._timeout -= 1
         self._timeout_text.set_text(self._timeout)
         if self._timeout == 0:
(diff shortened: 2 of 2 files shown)


quozl:
Thanks.  Reviewed.  Also reviewed uses of `TimeoutAlert` and I agree.  Please also;

- [ ] check no activities use `NotifyAlert`, as this is also based on `_TimeoutAlert`, and is far more frequently used?

- [ ] adjust `examples/alert.py` to report the timeout; and perhaps follow the style of the example code in `src/sugar3/graphics/alert.py`,

Fixes #366 (mentioned because you wrote "Fix #366" which did not trigger the automatic GitHub linking from the issue to the commit and pull request).

Pro-Panda:
@quozl, Thank you for pointing my mistake

 - [x] Checked all activities for `notifyAlert` as well, [here](https://github.com/search?p=1&q=org%3Asugarlabs+notifyalert&type=Code). (apologies, I missed it the first time)
   No activity uses the implementation that has been changed.
 - [x] Found 3 activities importing notifyAlert but not using it. (Referenced here and made PR's)
 - [ ] Changing the example to demonstrate the effect of timeout

quozl:
Thanks, well done.

Pro-Panda:
**Update:** Example modified to explain the various response states.

Pro-Panda:
[Github:Sugarlabs](https://github.com/sugarlabs/) does not contain all activity repositories, and I checked for use of `NotifyAlert` and `TimeoutAlert` only in repos under https://github.com/sugarlabs/
https://github.com/sugar-activities [presumably unofficial] is also not updated since 21/04/17

Should I write a script to download all extract all activities and get all Notify and Timeout Alert instances.

@quozl,
Kindly suggest.

quozl:
@Pro-Panda, thanks for the research.

activities.sugarlabs.org does contain activities for older releases of Sugar that no longer work with the most recent Sugar, and the Browse activity sends a Sugar version over HTTP to restrict access by known working version.

My preference is to abandon these activities that have no maintainer and no testing.  We have many activities in GitHub that don't work; plenty of work to do, so there's nothing to be gained by adding more.

If you have time enough to do that kind of analysis of activities.sugarlabs.org, might I instead suggest you test activities on Sugar or adopt activities for maintenance? :grin:

Pro-Panda:
@quozl, sure
I can adopt a few activities, priority to working on activities already on github.
But, I couldn't find a list of orphaned activities, do you have any in mind ?
Thanks

quozl:
I've no list.  Having a list would be the first step toward reversing the abandonment.

Pro-Panda:
Thank you for the idea. I will make a progressive list of orphaned activities on wiki by tonight

Resting my case; instances of `NotifyAlert` and `TimeoutAlert` in orphaned activities will be fixed (if needed) when maintainers adopt and test them

quozl:
Agreed.  You might use GitHub API to script;

- [ ] a list of activities; an activity has an `activity/activity.info` file in the repository,
- [ ] a list of activities with no recent commits; these may need fundamental maintenance such as testing, porting, merging translations, and releasing,
- [ ] a list of activities with commits after the most recent release tag; these may need releasing,
- [ ] a list of activities with branches or forks that are in advance of the `master` branch; these may need merging, upstreaming, or other actions.

Issue #378, opened 2018-02-07
suga3.activity.bundlebuilder dist_xo has no metadata warnings


quozl:
activity.info metadata is helpful for some services, especially activities.sugarlabs.org which requires several tags.

Change the `dist_xo` target to issue non-fatal warning messages when these tags are missing.

PR #380, opened 2018-02-11
Fix: #378 Issue non-fatal warnings on missing metadata

Pro-Panda:
**Fixed in this PR:** #378
**Changes made:**  
 - `python setup.py dist_xo` will now issue warnings(non-fatal), if `activity.info` won't have `license` or `activity_version` tags
 - `name` tag already issued a fatal error, if missing and hence was excluded.

**Suggestion:**
All tags independent of whether they are necessary for `aslo` should issue a warning. Kindly let me know, I will update the PR

**Tested**

### DIFF
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -237,6 +237,7 @@ class XOPackager(Packager):
     def __init__(self, builder):
         Packager.__init__(self, builder.config)
 
+        self.metadata_keys = ['activity_version', 'license']
         self.builder = builder
         self.builder.build_locale()
         self.package_path = os.path.join(self.config.dist_dir,
@@ -245,7 +246,7 @@ class XOPackager(Packager):
     def package(self):
         bundle_zip = zipfile.ZipFile(self.package_path, 'w',
                                      zipfile.ZIP_DEFLATED)
-
+        self.metadata_warn()
         for f in self.get_files_in_git():
             bundle_zip.write(os.path.join(self.config.source_dir, f),
                              os.path.join(self.config.bundle_root_dir, f))
@@ -256,6 +257,16 @@ class XOPackager(Packager):
                                           'locale', f))
 
         bundle_zip.close()
+
+    def metadata_warn(self):
+        with open(os.path.join(self.config.source_dir, 'activity','activity.info')) as f:
+            activity_metadata = f.readlines()
+        for key in self.metadata_keys:
+            for metadata in activity_metadata:
+                if metadata.startswith(key):
+                    break
+            else:
+                logging.warning('"' + key + '"' + ' key is missing in activity.info')
 
 
 class SourcePackager(Packager):
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -237,7 +237,6 @@ class XOPackager(Packager):
     def __init__(self, builder):
         Packager.__init__(self, builder.config)
 
-        self.metadata_keys = ['activity_version', 'license']
         self.builder = builder
         self.builder.build_locale()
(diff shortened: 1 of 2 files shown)


quozl:
You've added to the `XOPackager` class.  I hadn't thought of that.

Better would be to check `config` in `cmd_dist_xo`, as at that point the file is already parsed and available.

Another place would be `class ActivityBundle` which will help with `aslo-v3` workflow which (at one stage) proposed not to require use of the `dist_xo` target by developers.

Regarding what tags must be present;

* compare against our toolkit's [authoritative specification of metadata keys](https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/src/sugar3/bundle/__init__.py#L20).
* search the in-production `aslo` repository for more details, see https://github.com/sugarlabs/aslo/search?utf8=%E2%9C%93&q=activity_version&type=
* search the in-development `aslo-v3` repository for more details, see https://github.com/sugarlabs/aslo-v3/search?utf8=%E2%9C%93&q=activity_version&type=
* see our [Wiki specification of the file format](https://wiki.sugarlabs.org/go/Development_Team/Almanac/Activity_Bundles#.info_file_format).

Pro-Panda:
Thank you for the detailed review and guidance


**Update:**
 - Decided to go with `ActivityBundle`, as it might help with `aslo-v3` workflow in future
 - Displaying warnings only for the mandatory tags and not the optional ones
 - Also displaying warnings for tags mandatory for AppStream

**Sources for looking up tags:** (Thanks to @quozl)
 - [ASLO release.py L141](https://github.com/sugarlabs/aslo-v3/blob/272b779725cd0ca45e8008b2514e051c9b7ca1d8/aslo/api/release.py#L141)
 - [/Bundle/__init__.py](https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/src/sugar3/bundle/__init__.py#L20)
 - [Wiki](https://wiki.sugarlabs.org/go/Development_Team/Almanac/Activity_Bundles#.info_file_format)

quozl:
Thanks.  Reviewed.

- [ ] there is a check for `name`, `bundle_id` and some other keys already in `_parse_info`,
- [ ] does not seem to be a good reason for a new method of the class, especially also as the method is not privately named; place it inline in `_parse_info` instead,
- [ ] other messages use "does not" rather than "doesn't"; please keep the messages consistent,

I'm not so sure about the AppStream changes, as there are almost no activities that have AppStream metadata (e.g. only Browse), so we would end up with lots of warnings in logs as Sugar starts up for no real benefit.  I don't think anybody is using our AppStream metadata anyway.  Anyway, I did notice;

- [ ] `Appstream` should be `AppStream` to be consistent with other uses,
- [ ] some of the your AppStream additions become redundant with `_generate_appdata` in `bundlebuilder.py`,

I've also updated the specification https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/7e216a04d579d6f0e8283e5d3975bf72edd2f67b.

Pro-Panda:
@quozl Thanks once again for the detailed review

**Changes made:**
 - [X] Removed redundant checks already checked in `_parse_info`
 - [X] placed the checks inline in `_parse_info`, rather than a new method
 - [X] Changed warning message, to maintain consistency
 - [X] Checks for AppStream metadata removed, considering that there are redundant checks and the 

Changes in 7e216a0 are for optional metadata. Kindly let me know if checks for optional metadata should also be added.

Kindly review
If no more changes are to be made, shall I squash all commits.

quozl:
Thanks.  Reviewed.

Rebased.  Rewrote messages for consistency with others.  Moved some checks into else clauses where the same metadata key was tested.

https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/cc2357c787cb30d02e1568a2bf23dc18c367468f

PR #381, opened 2018-02-18
Fix: Hide notebook(if present) on fullscreen

Pro-Panda:
**Fixed in this PR:** [Terminal Activity #25](https://github.com/sugarlabs/terminal-activity/issues/25)

**Effect of this PR:**
 - fullscreen will hide notebook tabs(if present)
 - unfullscreen will show notebook tabs

@quozl,
I am in doubt whether this change should have been implemented in the toolkit, or local to the terminal activity.
My opinion: Fullscreen implies the current working space should be the only thing present on the entire screen. Hence, this change should be there for all activities, in the toolkit itself

**Tested on:** 
terminal activity, Sugar 0.112
(rdesktop, Ubuntu 16.04)

### DIFF
--- a/src/sugar3/graphics/window.py
+++ b/src/sugar3/graphics/window.py
@@ -98,7 +98,8 @@ class Window(Gtk.Window):
 
     Used as a container to display things that happen in an activity.
     A window must contain a canvas widget, and a toolbar box widget.
-    A window may also contain alert message widgets and a tray widget.
+    A window may also contain alert message widgets, tray widget and
+    notebook widget.
 
     Widgets are kept in a vertical box in this order;
         * toolbar box,
@@ -140,6 +141,7 @@ class Window(Gtk.Window):
         self._alerts = []
         self._canvas = None
         self.tray = None
+        self._notebook = None
 
         self.__vbox = Gtk.VBox()
         self.__hbox = Gtk.HBox()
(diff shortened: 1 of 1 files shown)


Pro-Panda:
If these changes are approved, kindly let me know before merging, as the following changes also need to be done.

 - [ ] checking all activities to see if this change can affect their intended implementation
 - [ ] If found, make changes in those activities, or in the implementation here.
 - [ ] making necessary changes in the developer docs
 - [ ] Adding notebook in this [list](https://github.com/Pro-Panda/sugar-toolkit-gtk3/blob/5d7f4b513ca94342ea3b8329eed1a9400120549d/src/sugar3/graphics/window.py#L104). I am not sure of the order.

quozl:
@Pro-Panda, thanks for the research.  It does begin to look complicated.  I've solved it in Terminal by adding methods to the activity class which use the superior methods.   https://github.com/sugarlabs/terminal-activity/commit/b37d83ff2f16717cb3dad08590f3c3c56f774232

Pro-Panda:
@quozl, Thank you 😄 
But shouldn't this be a global feature, pertaining to all activities ?

Eg: [pydebug](https://github.com/sugarlabs/pydebug-activity) also has a fullscreen option, which doesn't hide the notebook (although tabs are at the bottom, and don't overlap with the unfullscreen button, but tabs should be hidden in fullscreen)

quozl:
Thanks.  I've not used or shipped pydebug.

I'm not sure if it is justified to make it global; there are so few activities that embed a Gtk.Notebook as activity canvas.

https://github.com/search?utf8=%E2%9C%93&q=org%3Asugarlabs+set_canvas%28self._notebook%29&type=Code is a useful search.

Activities that I'm caring for that have a Gtk.Notebook;
* Browse; yet to be fixed, https://github.com/sugarlabs/browse-activity/issues/69
* Speak; does not use tabs,
* Terminal; now fixed,
* Pippy; does not have a fullscreen button.

Pro-Panda:
Thank you for the explanation. 
I too agree now, that making it global is not justified

My mistake was looking just for instances of Notebook ([here](https://github.com/search?utf8=%E2%9C%93&q=org%3Asugarlabs+notebook&type=Code)) before opening the PR.

Issue #382, opened 2018-02-26
Port to Six


quozl:
Port the Sugar Toolkit for GTK+ 3 to Six, providing both Python 2 and Python 3 bindings from this package.

Progress goals:

- [x] fix a critical accidental dependency; remove the `from jarabe.model import shell` from `src/sugar3/graphics/popwindow.py` and use some other design,
- [x] <s>port C code in `src/sugar3`</s> _none of the C files include Python,h_,
- [x] port toolkit code in `src/sugar3` which calls C code, and test that the C code can be called from REPL,
- [x] port toolkit code in `src/sugar3/graphics`,
- [x] port graphics examples, and test that examples work,
- [x] port toolkit code in `src/sugar3/activity`,
- [x] port the [Hello World activity](https://github.com/sugarlabs/hello-world-fork),
- [x] port all remaining toolkit code, such as `src/sugar3/datastore`, `presence`, `bundle`,
- [x] port to TelepathyGLib,
- [x] port make files, and test that "make install" installs both bindings,
- [x] regenerate documentation,
- [ ] <s>make local Fedora and Debian packages for testing by others.</s>
- [X] release new version of toolkit.

_this is an issue for tracking a major project which is larger than one person's contributions, so developers should make pull requests to satisfy any of these items, and not ask to own the issue._

yashagrawal3:
@quozl I would like to work on this issue

quozl:
A discussion began in pull-request https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/385 about redundant work.  As your release manager, my opinion is;

* everyone who contributes to issue https://github.com/sugarlabs/sugar-toolkit-gtk3/issues/382 should cherry-pick each other's patches into a local branch before starting any work, so that you can avoid doing work that was already done.
* you should all learn how to and be comfortable enough to maintain several branches within your own local clones.
* when making a pull request, rebase your changes to master.
* a pull request may contain work by you alone, or it may contain patches by others too.
* a pull request may be for review only, or for review and merging.
* to aid in review and collaboration, you may use a branch name of _python3_.

I'm unwilling to merge any pull requests until the whole issue is resolved; as I don't want master branch in a broken state in any point of git history, because this breaks use of "git bisect".

quozl:
[...]

PR #383, opened 2018-03-01
Port to Six

pro-panda:
**Changes made/being made in this PR:** #382


Pro-Panda:
**Note:** My current method of testing is by trying `python3 <filenmae>` and removing any errors that come up. Please suggest if there are better testing methods as this point of port

Pro-Panda:
**Note:** My current method of testing is by trying `python3 <filename>` and `python2 <filename>` and removing any errors that come up. Please suggest if there is a better alternative as this point of porting

Pro-Panda:
@walterbender @quozl 
Kindly suggest improvements in my initial approach.

**Doubts:**
 - telepathy doesn't exist for python3. Can we use [this](https://github.com/Pro-Panda/telepathy3), or should I look for an alternative
 - Kindly guide on what approach to follow to remove `from jarabe.model import shell` from `src/sugar3/graphics/popwindow.py`

walterbender:
It will be hard to test the individual files stand-alone and I don't think our build system tests will be of much use. I will think about it.

Re telepathy, we'll have to do extensive testing of the Python 3 port as well. But maybe we can also ping Collabora to see if they have a Python3 equivalent (more likely to be maintained) toolkit we can begin migrating towards.

Pro-Panda:
Re: telepathy. I agree, it needs to be tested extensively, and so does the toolkit ports. Until Collabora responds, I will either avoid `telepathy` ports or use the port I made.

Another doubt:
I went throught the `extensions` (C code). Kindly guide that what componnents will be affected and will need to be ported ?

quozl:
I don't know how to fix the reference to `jarabe.model`.  Can the needed functionality be moved into the toolkit?

Who is pinging Collabora?

I don't know which `extensions` will be affected.

Pro-Panda:
@quozl 
Thank you for letting me know
> Moving the jarabe.model functionality to the toolkit

There are a lot of interdependencies, but I'll try to do it.
> Who is pinging collabora

I misread @walterbender  message, Sorry. Edited my response

I will try to see which `extensions` might require changes

quozl:
Reviewed up to 743742d.

quozl:
Reviewed to 53f6a20.

Pro-Panda:
**Update on the PR:**
 - All `python2` code has been ported to `six`
 - Tested few examples, no issues found

Pro-Panda:
**Update on the PR:**
 - All `python2` code ported to `six`
 - Tested all examples, on both `python2` and `python3` [Logs: [here](https://gist.github.com/Pro-Panda/af8382c343cea38052d27046391e333f)]
 - Ran sugar using this toolkit. Tested couple of activities, journal and other features. No issues experienced. Logs also didn't show errors attributing to the `toolkit`

**Work in progress:**
 - Making fix for issues highlighted by the logs
 - Figuring a workaround for `src/sugar3/graphics/popwindow.py` import from `jarabe`

quozl:
Reviewed to 89dc9ed.

Pro-Panda:
@quozl,
I am using `sugar-activity` and `sugar-activity3` for running activities on `python 2` and `python 3` 

A `python 3` activity must have the `python_version` tag in `activity.info`
(Atleast until we have completed porting all activities)

Do you think there might be a better alternative than this.

**Note:** Tested the changes with the `python 3` port of CowBulls activity, [here](https://github.com/Pro-Panda/CowBulls-activity/tree/python3)

quozl:
Reviewed to 9699941.

I do not think there will be a port of all activities.  It would be nice, but I'm a realist.  :grin:

Perhaps a better alternative is to use in activity.info;
```
exec = sugar-activity3 activity.ClassName
```

Also consider;

* documentation for `exec` in `src/sugar3/bundle/__init__.py`
* `tests/data/sample.activity/activity/activity.info`

Now that you have a ported activity, please also contribute to https://github.com/sugarlabs/sugar-docs/issues/145.

quozl:
Reviewed to a50e843.

quozl:
Reviewed to 19b551f.  These last two deserve to jump the queue and land now, but 19b551f has several merge conflicts.  Your ccbf8a475c55e592bbc1dd66260e9a6c97b15ab7 has been cherry-picked and pushed as 8f047b58195dfed88f4cf4aef3ac122ccd5ff0eb.

quozl:
Reviewed to 1730e21.

doc/index.rst
```diff
@@ -2,7 +2,7 @@ Sugar Toolkit GTK3
 ==================
 
 Sugar Toolkit GTK3, or `sugar3`, is a toolkit for writing Sugar
-activities in Python:
+activities in Python 3:
```
quozl: Also in Python 2, so I suggest removing this change.

configure.ac
```diff
@@ -15,7 +15,7 @@ GNOME_COMPILE_WARNINGS(maximum)
 
 AC_PATH_PROG([GLIB_GENMARSHAL], [glib-genmarshal])
 
-PYTHON=python2
+PYTHON=python3
```
octamois: It specifically installs for python3 
We need toolkit to be 2&3 compatible

Pro-Panda:
@quozl, no.
The package should ideally install for both Python 2 and Python 3. I am still working on how to do that

configure.ac
Pro-Panda: yes, I know. I don't yet have a fix for installing for both Python 2 and Python 3.

I am leaving this change as is for now, since LEM doesn't hold here.
Reference: https://en.wikipedia.org/wiki/Law_of_excluded_middle

quozl:
We need the source package (the .tar.xz) to be capable of being installed for both python2 and python3, but not at the same time.  The duty to do both is given to the user of the source package; such as the Fedora or Debian packager, or an end-user doing a native install by hand.

So I'll consider the changes for sugar-toolkit-gtk3 to be complete in this respect once PYTHON can be set to either, at the shell prompt.  At the moment, we're overriding PYTHON rather than using a default.

Pro-Panda:
@quozl 
[examples/ticket2855](https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/examples/ticket2855.py), ticket2999 and ticket3000 don't seem to be linked to https://bugs.sugarlabs.org/ticket/2855, and others respectively. Is there a different ticket they are referring to ?

quozl:
> Is there a different ticket they are referring to ?

Yes, the OLPC bug tracker; e.g. http://dev.laptop.org/ticket/2855

quozl:
Reviewed 1f34c2d.

I don't understand why 57168c7 is needed; why can't a Sugar Web activity, which is written in JavaScript, use `sugar-activity-web` ported to Python 3?

quozl:
Reviewed 35670c9.

quozl:
Merge conflict was introduced by 13a2282e5e179b243281de3586bb227114a534d5 and following commits on *master* branch.  Did cherry-picks and push.

quozl:
Rebased on *master*.

quozl:
Reviewed to cc3050f.

quozl:
Reviewed to bd386f7.

quozl:
Reviewed to 0ec673f.

quozl:
@pro-panda, I'm unable to `import sugar3` in Python 3.  Have you any idea why?

Here's what I've done;
* installed a Debian Stretch VM, using Sugar Live Build,
* checked out this branch, `autogen.sh`, `configure`, `make`, `make install`,
* tried `import sugar3` within Python, which gives `ImportError: No module named 'sugar3'`,

Here's what I've found since;
* the `make install` has copied files to `/usr/lib/python3.5/site-packages/sugar3`,
* installing `python3-serial` package has placed files in `/usr/lib/python3/dist-packages`,
* renaming `site-packages` to `dist-packages` fixes the import fail,
* moving the `sugar3` directory from the first to the second directory fixes the import fail.

So it would seem that our `make install` uses the wrong place.

pro-panda:
Thanks for testing. I agree. Reproduced on Ubuntu 16.04.

ritjoe:
On Sun, Sep 23, 2018 at 08:11:36PM -0700, James Cameron wrote:
> @pro-panda, I'm unable to import sugar3 in Python 3. Have you any idea why?
> 
> Here's what I've done;
> 
>   • installed a Debian Stretch VM, using Sugar Live Build,
>   • checked out this branch, autogen.sh, configure, make, make install,
>   • tried import sugar3 within Python, which gives ImportError: No module named
>     'sugar3',
> 
> Here's what I've found since;
> 
>   • the make install has copied files to /usr/lib/python3.5/site-packages/
>     sugar3,
>   • installing python3-serial package has placed files in /usr/lib/python3/
>     dist-packages,
>   • renaming site-packages to dist-packages fixes the import fail,
>   • moving the sugar3 directory from the first to the second directory fixes
>     the import fail.
> 
> So it would seem that our make install uses the wrong place.

This is just a start as to how to understand how to reconcile the two locations:

https://stackoverflow.com/questions/9387928/whats-the-difference-between-dist-packages-and-site-packages#9388115

but from that it looks like each method of installing was doing what it was supposed to do, under Debian policy.

The question is how to reconcile the different uses of the two?

Rather than trying to make one or the other work, perhaps configuring path variables to check both types of locations is the way forward?

quozl:
I'm guessing from the policy Debian just doesn't support user-installed system-wide Python packages without the user also changing the path.

Best ways forward seem to depend on target;
* for Sugar Live Build, which does a `git clone` then `make install`, we'll have to move the `sugar3` directory as a workaround,
* for Debian or Ubuntu packaging building, no action needed.

I've not evaluated Fedora.

pro-panda:
> for Sugar Live Build, which does a git clone then make install, we'll have to move the sugar3 directory as a workaround,

I agree. I couldn't find an alternate as well

quozl:
@pro-panda, tested on Ubuntu 18.04 and fixed a problem with starting activities, see d3ac3e5.

quozl:
Merged as aa8a5e70c415e6c2acf4ff373d9b366ac4692bb1, thanks!

Aniket21mathur:
Any further work required to be done on this pr? It is still a part of Project Task Checklist for port to python3 project. Thanks!

quozl:
See https://github.com/sugarlabs/sugar-toolkit-gtk3/issues/382 for the work remaining to be done.

configure.ac
Aniket21mathur: @pro-panda so the sugar toolkit is not yet completely 2 & 3 compatible?

quozl: It is compatible.  You can override PYTHON when you call configure.ac, right?

Aniket21mathur: Yes @quozl , as per my undersatnding in that case it will install either python 2 or python 3 and it totally depends upon user how they want to install the toolkit.

Aniket21mathur:
> I'm guessing from the policy Debian just doesn't support user-installed system-wide Python packages without the user also changing the path.
> 
> Best ways forward seem to depend on target;
> 
> * for Sugar Live Build, which does a `git clone` then `make install`, we'll have to move the `sugar3` directory as a workaround,
> * for Debian or Ubuntu packaging building, no action needed.
> 
> I've not evaluated Fedora.

I think there is a need to include this in the Native build documentation. Suggestions?

quozl:
Sure.  It's a mess.  Debian and Fedora both seem to have problems dealing with installing Python programs using `autotools` into `/usr/local`.  It is a path not often tread, so it is unlikely to gain attention.

Sub task _make local Fedora and Debian packages for testing by others_ in each of our Port to Python 3 issues should help to ensure our release, when made, will build and run on Fedora and Debian.

Issue #375, opened 2018-01-12
sugar3.activity.widgets.TitleEntry ignores enter key


quozl:
Learners learned to use `tab` to finish editing TitleEntry, followed by clicking the activity icon to dismiss the dialog.  A slow work flow and inconsistent with journal list view.  The `enter` key is ignored.

- [x] `enter` key must finish editing,
- [ ] `enter` key should dismiss the dialog on activities that use one (most of them except Chat),
- [ ] test with or without the save-as feature,
- [ ] backport to https://github.com/sugarlabs/sugar-toolkit.

Arose from problem report https://github.com/sugarlabs/sugar/issues/771

Pro-Panda:
@quozl I would like to work on this issue

quozl:
https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/f3dd2c2042ad5b3fd5256838c0ec201da9402406 fixes "enter key must finish editing".

chimosky:
@quozl while taking a look at this, it seems  `enter key should dismiss the dialog on activities that use one (most of them except Chat)` already happens, it looks like there's no way to set the save-as feature from within sugar without using gsettings.

chimosky:
@quozl I'll be closing this as the issue no longer exists.

PR #384, opened 2018-03-02
[Fix]: #375 (sugar3.activity.widgets.TitleEntry ignores enter key)

Pro-Panda:
**Fixed in this PR:** #375 
- [X] `enter` key should dismiss the dialog on activities that use one (most of them except Chat),
- [x] test with or without the save-as feature,
- [ ] backport to https://github.com/sugarlabs/sugar-toolkit.

**Note:** This is an in progress PR

**Tested on:**
Sugar 0.112, Ubuntu 16.04

@quozl Kindly review

### DIFF
--- a/src/sugar3/activity/widgets.py
+++ b/src/sugar3/activity/widgets.py
@@ -19,6 +19,7 @@
 
 from gi.repository import Gdk
 from gi.repository import Gtk
+from gi.repository import GObject
 import gettext
 
 from sugar3.graphics.toolbutton import ToolButton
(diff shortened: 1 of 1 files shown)


Pro-Panda:
I couldn't find an example activity using save-as feature in title entry.
I will do a backport after testing with a save-as feature.

quozl:
Reviewed cd4670f.

To test with save-as, you must enable save-as and restart Sugar.  One activity that has an awareness of the save-as feature is Browse, which calls `profile.get_save_as`.  Other activities usually don't need an awareness; save-as is implemented in the toolkit.

Pro-Panda:
**Update:** Tested the `save-as` feature on Browse.

Pro-Panda:
Fixes #375

PR #386, opened 2018-03-26
sugar3.speech - remove GConf compatibility

quozl:
GConf compatibility support in sugar3.speech module allows activities to read or be notified of changes to the speech pitch and rate.  GConf is deprecated.

Remove GConf from the sugar3.speech module.

The speech pitch and rate will no longer be available via GConf.  Fix GTK+ 2 activities by porting to GTK+ 3.  Fix GTK+ 3 activities by porting from GConf to Gio.Settings.

Tested on Ubuntu 16.04.

### DIFF
--- a/src/sugar3/speech.py
+++ b/src/sugar3/speech.py
@@ -288,11 +288,6 @@ class SpeechManager(GObject.GObject):
 
     def save(self):
         self._save_timeout_id = -1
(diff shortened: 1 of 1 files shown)


quozl:
Pushed as 30139a3a9ed444c365c5c5d98e6bafde5b6c711d.

PR #391, opened 2018-05-20
Remove unwanted dependency of toolkit to jarabe

pro-panda:
- Add exmaple demonstrating `sugar3.graphics.popwindow.PopWindow`
 - Remove dependancy of `sugar3.graohics.popwindow` to `jarabe`

### DIFF
--- a/examples/popwindow.py
+++ b/examples/popwindow.py
@@ -1,0 +1,16 @@
+from sugar3.graphics.popwindow import PopWindow
+
+import common
+test = common.Test()
+
+
+def popwindow_closed(widget):
+    print("Popwindow closed")
+
+
+c = PopWindow()
+c.connect('destroy', popwindow_closed)
+c.show_all()  # Press `escape` key or click on the close_button to close the popwindow
+
+if __name__ == '__main__':
+    common.main(test)
--- a/src/sugar3/graphics/popwindow.py
+++ b/src/sugar3/graphics/popwindow.py
@@ -35,8 +35,6 @@ from gi.repository import GObject
 from sugar3.graphics import style
 from sugar3.graphics.toolbutton import ToolButton
 
-from jarabe.model import shell
-
 
 class PopWindow(Gtk.Window):
     """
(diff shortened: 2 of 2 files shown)


Pro-Panda:
Tested on Ubuntu 16.04, Sugar 0.112

Pro-Panda:
I couldn't find what to fix in `sugar:src/jarabe/desktop/activitychooser.py`.

Didn't understand 'window remains modal' :sweat_smile:

Pro-Panda:
Sorry, but I don't think this change removed the Modal behaviour. I think I misunderstood

quozl:
I'll explain further.

When the Journal project feature was added, one of the new dialogs was an activity chooser; which displays a list of activities.  As most dialogs in Sugar are modal, so was this new dialog.  Part of the implementation of this dialog was placed in the `sugar-toolkit-gtk3` module `sugar3/graphics`, and part in the `sugar` shell `jarabe`.

Part of the modal behaviour is the tracking of multiple modal dialogs at the same time.  Such as <kbd>alt</kbd>+<kbd>shift</kbd><kbd>m</kbd> which launches My Settings.  That's what `push_modal` and `pop_modal` do.

In removing the modal behaviour tracking from the toolkit in your commits here, you effectively remove the modal behaviour tracking from the dialog in the shell.

So can you either show a test case to demonstrate that modal behaviour tracking is not required, or move the calls into the shell?

I'm trying out the Journal project feature at the moment to try to make such a test case, but I've forgotten how it works and it doesn't seem to be working the way I remember.

Pro-Panda:
Thank you for the detailed explanation. 

I tested using [Sugar Toolkit @58224cb](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/58224cb15d84765e79492c7bb513e952e856cbfe) and [Sugar @489974f](https://github.com/sugarlabs/sugar/commit/489974f4243eb3a18c0bea07812c436ed8aa7d5c), and the behaviour of `ActivityChooser` is not modal. 

I am able to access other sections outside of the `PopWindow`. I'll try to make a patch in `jarabe` to fix the same, thanks

Pro-Panda:
While I am now working on making a patch for `jarabe`, I think this can be merged since `ActivityChooser` was not modal before these changes

Pro-Panda:
Noted, thanks

pro-panda:
Closing. Relevant PR open at https://github.com/sugarlabs/sugar-toolkit-gtk3

PR #394, opened 2018-05-30
Define new parameters for activity metadata

vipulgupta2048:
With reference to https://github.com/sugarlabs/activity-abacus/pull/17, new parameters are needed to be defined in activity metadata [activity.info] as other activities have been using them in reference to activity.info of [turtleart-activity](https://github.com/sugarlabs/turtleart-activity/blob/master/activity/activity.info) and others. 
This is important for displaying correct and accurate information of the activity on the web portal. 
The new parameters being:-    
1. category/categories 
2. website 

The description has also been updated for the same. All suggestions and edits are appreciated. 
- With respect to this, the example has also been updated of the `browse` activity that is listed below. 
- I have also taken the liberty to improve the description of already defined parameters from this [1](https://wiki.sugarlabs.org/go/Development_Team/Almanac/Activity_Bundles#.info_file_format). Hence, request the maintainers to depreciate the aforementioned page in favor of this list that is present here.
- I have also tried to maintain flake8, but I apologize if at someplace I had to wing it. 
 
**Questions**
Upon comparison of the content from [1](https://wiki.sugarlabs.org/go/Development_Team/Almanac/Activity_Bundles#.info_file_format) with [2](https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/src/sugar3/bundle/__init__.py)
1. I found the `exec = sugar-activity webactivity.WebActivity -s` in the example of the browse activity having an argument `-s`. Is this functionality deprecation if not we should add the same to 2. 
2. Wiki page [1](https://wiki.sugarlabs.org/go/Development_Team/Almanac/Activity_Bundles#.info_file_format) still contains a lot of detailed description of parameters not listed in 2, since we are going ahead with the depreciation of 1. I suggest we transfer them to 2, as developers will now be referring this.

### DIFF
--- a/src/sugar3/bundle/__init__.py
+++ b/src/sugar3/bundle/__init__.py
@@ -21,7 +21,7 @@ Activity Metadata
 =================
 
 Your `activity/activity.info` file must have these metadata keys after
-an `[Activity]` header:
+an `[Activity]` header on the first line:
 
 * `name` - the name of the activity, shown by Sugar in the list of
   installed activities, e.g. Browse,
@@ -46,28 +46,37 @@ an `[Activity]` header:
 
 Optional metadata keys are;
 
-* `summary` - a short summary of the activity that may be displayed in the List or Home Views,
+* `summary` - a short summary of the activity that may be displayed in
+  the List or Home Views,
 
 * `mime_types` - list of MIME types supported by the activity,
   separated by semicolons.  Your `read_file` method must be able to read
   files of these MIME types.  Used to offer your activity when opening a
   downloaded file or a journal object.
 
-* `url` - link to the home page for the activity,
+* `url` - link to a home page or user documentation on
+  https://help.sugarlabs.org/,
 
 * `repository` - link to repository for activity code, for use by git clone,
 
 * `single_instance` - if yes, only a single instance of an activity
-  should be started, and if another instance is requested the existing
-  instance shown,
+  should be started at any one time, and if another instance is requested
+  the existing instance shown,
 
-* `max_participants` - maximum number of participants for sharing an activity,
+* `max_participants` - maximum participants for sharing an activity,
 
-* `tags` - more context into which to place an activity,
+* `tags` - a semicolon or whitespace delimited list of keywords that
+   describe the activity.  Suggested keywords are Programming,
+   Robotics, Internet, Science, Maths, Language, Geography, Documents,
+   Music, Media, Art, Teacher, or System,
 
 * `show_launcher` - if set to "no", the activity is not shown in list view,
 
 Deprecated metadata keys are;
+
+* `category` or `categories` - aliases for `tags`,
+
+* `website` - alias for `url`,
 
 * `update_url` - the updater no longer uses this.
 
@@ -103,7 +112,6 @@ Example `activity.info`
 -----------------------
 
 .. code-block:: ini
-    :emphasize-lines: 10-12,20-21
 
     [Activity]
     name = Browse
@@ -115,6 +123,10 @@ Example `activity.info`
     summary = Surf the world!
 
     license = GPL-3.0+
+    repository = https://github.com/sugarlabs/browse-activity.git
+    url = https://help.sugarlabs.org/en/browse.html
+    tags = Utilities;Web
+
     metadata_license = CC0-1.0
     description:
         <p>Surf the world! Here you can do research, watch educational videos, take online courses, find books, connect with friends and more.  Browse is powered by the WebKit2 rendering engine with the Faster Than Light javascript interpreter - allowing you to view the full beauty of the web.</p>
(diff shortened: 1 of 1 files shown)


Pro-Panda:
> I found the exec = sugar-activity webactivity.WebActivity -s in the example of the browse activity having an argument -s. Is this functionality depreciated if not we should add the same to 2.

I don't think it should be added to `bundle/__init__.py`. It does not discuss any of the arguments of sugar-activity. 

For reference, they are added [here](https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/bin/sugar-activity#L84)

quozl:
@vipulgupta2048, thanks for the patch.  Please use the word `deprecated` instead of `depreciated`.  To deprecate means to express disapproval of.  Oxford dictionary has a pertinent usage;

> _be deprecated_ (chiefly of a software feature) be usable but regarded as obsolete and best avoided, typically because it has been superseded.

Rather than deprecate, the content of the section on the Wiki page should be deleted and replaced with a link to developer documentation https://developer.sugarlabs.org/sugar3/sugar3.bundle.html#submodules and the source `bundle/__init__.py`.

I'll review the patch next.

src/sugar3/bundle/__init__.py
```diff
@@ -46,25 +47,31 @@
...
-* `summary` - a short summary of the activity that may be displayed in the List or Home Views,
+* `summary` - a short summary of the activity that may be displayed in
+  the List or Home Views,
 
 * `mime_types` - list of MIME types supported by the activity,
   separated by semicolons.  Your `read_file` method must be able to read
-  files of these MIME types.  Used to offer your activity when opening a
-  downloaded file or a journal object.
+  files of these MIME types.  Used to offer your activity in a list of activities
+  when opening a downloaded file from the web or a certain journal object.
```
quozl: or from a USB drive, or by peer to peer file transfer, ... mentioning all the ways a journal object can be created seems a bit too much, and detracts from the core message.

src/sugar3/bundle/__init__.py
```diff
@@ -46,25 +47,31 @@
...
+  should be started at any one time, and if another instance is requested
+  the existing instance shown,
 
-* `max_participants` - maximum number of participants for sharing an activity,
+* `max_participants` - maximum number of participants that can join a shared activity,
 
 * `tags` - more context into which to place an activity,
 
+* `category`/`categories` - define/choose a category/categories that fits your
+activity suitably. You can choose from the list available at http://activities.sugarlabs.org/en-US/sugar/,
```
quozl: When you use <kbd>/</kbd> to separate terms, the meaning is ambiguous; it could mean and, or, or and/or.

Add a list of category names here rather than refer to an obsolete web service.

Define the format of the value; space or <kbd>;</kbd> delimited.

My preference is to use `tags`, with `category` and `categories` to be acceptable aliases.

src/sugar3/bundle/__init__.py
```diff
@@ -46,25 +47,31 @@
...
+
+* `website` - link to activity-documentation as present in help-activity,
```
quozl: We already have `url`, so you are adding an alias for `url`, right?

Perhaps additionally describe `url` as "or link to the activity on https://help.sugarlabs.org/ "

vipulgupta2048: Yes, in the case if the activity already has an `url` defined, I didn't want to overwrite anybody's else work. Additionally described `url` as well.

vipulgupta2048:
Changes made as requested. 
- Category list is taken from https://aslo3-devel.sugarlabs.org
- Categories will use delimiter as `;` - Also demonstrated the same in help-activity.  
- Also, I noticed the example of the browse activity missing some details. Mainly, 
1. `mime_types = image/png;image/gif;image/jpeg;text/html;text/uri-list;application/xhtml+xml;application/rss+xml;application/xml;text/css
`
2. `single_instance = no
`

quozl:
> Changes made as requested.

Not quite right, but I've no time to explain right now.  I may push something tomorrow if you haven't fixed it.

> noticed the example of the browse activity

https://github.com/sugarlabs/browse-activity/blob/master/activity/activity.info has a much longer `mime_types`.  I think it is sufficient to abbreviate and use an ellipsis.

`single_instance` is optional and not in Browse, so I don't think it should be in the example.

quozl:
Made changes rather than explain in detail.  Pushed to your _master_ branch as ca71715.  Please review [aggregate change](https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/394/files) (or `git pull && git diff origin/master`) and comment.

For future pull requests, please work in a branch other than _master_.  Your use of _master_ made contributing more difficult for me, and will make it more difficult for you.  See our guide to [Contributing](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md) and search in page for _branch_ to get an idea of why we recommend not using _master_.

vipulgupta2048:
@quozl I see your changes and they LGTM. I see the point you were trying to make and understand the need for your changes rather than explaining. 

>For future pull requests, please work in a branch other than master.

I will surely keep that in mind all my future PR's

quozl:
Good, so we are both in agreement; now, any other reviewers for the aggregate change?  @Pro-Panda, @walterbender, @yashagrawal3?

* encourage `url` for either home page or user documentation,
* give examples for categorisation tags, define format,
* make `website` an alias for `url`,
* make `category` and `categories` aliases for `tags`.
* add the aliases to deprecated list; valid for use, but not encouraged.

quozl:
Squashing.

quozl:
Edited the Wiki page, removing what is in the `bundle/__init__.py` file, leaving historical and additional information not suited to being in git.

vipulgupta2048:
Thanks @quozl

PR #395, opened 2018-05-30
fix datastore status signal name usage

Pro-Panda:
This pull request _will_ incorporate changes needed/observed while working on [Datastore/#8](https://github.com/sugarlabs/sugar-datastore/pull/8)

### DIFF
--- a/src/sugar3/datastore/datastore.py
+++ b/src/sugar3/datastore/datastore.py
@@ -58,7 +58,7 @@ def _get_data_store():
 
 def __datastore_created_cb(object_id):
     metadata = _get_data_store().get_properties(object_id, byte_arrays=True)
-    updated.send(None, object_id=object_id, metadata=metadata)
+    created.send(None, object_id=object_id, metadata=metadata)
 
 
 def __datastore_updated_cb(object_id):


quozl:
Fascinating, thanks.  I don't have time right at the moment to test, but might your change break the following test case?

* start Sugar, empty journal, check that the activity icons are without user colour,
* start an activity,
* return to the home view and check that the activity icon has changed to user colour,
* quit the activity,
* verify the activity icon has changed to user colour.

The shell connects to the `updated` signal and not the `created` signal.  I'm not sure if there will be a `created` followed by `updated`.  An activity that creates an object but does not write to it might only cause `created`.  Here's the shell code;

https://github.com/sugarlabs/sugar/blob/master/src/jarabe/desktop/favoritesview.py#L430

I found no activities that use this `updated` signal.  Some activities use the D-Bus API.

The history of this typo is interesting;
* added in https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/e1d4afe13e3b49f01083a00ca4b83cc4011829d8
* based on work in bug https://bugs.sugarlabs.org/ticket/1197
* first patch draft https://bugs.sugarlabs.org/attachment/ticket/1197/sugar-toolkit-0001-use-sugar.datastore.datastore-or-at-least-sugar.data.patch has no typo,
* second patch draft https://bugs.sugarlabs.org/attachment/ticket/1197/0001-use-sugar.datastore.-instead-of-direct-DBus-access.patch has no typo, but shows changes to arguments in the functions,
* merged patch http://git.sugarlabs.org/sugar-toolkit/mainline/commit/e1d4afe13e3b49f01083a00ca4b83cc4011829d8 has the typo, consistent with a copy and paste error while changing arguments.

Pro-Panda:
Agreed. My change fails this test case.

- start Sugar, empty journal; the activity icons are without user colour,
- start an activity,
- return to the home view; the activity icon color doesn't change,
- open the activity from the home view; a new instance is created,
- return to the home view; activity icon color changes

While the `shell` uses `updated`, I don't see `created` being used anywhere. Maybe it should be removed ?

Out of context, but isn't [this](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/da678c6973c124c81f2365f93e0df6d3f93d995f#diff-1f10da030d5249a0fef9217d287dc2efR178) warning redundant, since we call destroy in the next line itself. Maybe I am confusing it ?

quozl:
> maybe it should be removed?

My preference is to merge your change on condition that you also connect to the `created` signal in https://github.com/sugarlabs/sugar/blob/master/src/jarabe/desktop/favoritesview.py#L430

>  isn't this warning redundant, since we call destroy in the next line itself.

The warning is important, as it shows a mistake was made by the caller.  We oblige the caller to fix their code.

PR #397, opened 2018-07-08
use sugar-activity-build in favor or adding setup.py to every activity

pro-panda:
Instead of using
```python
./setup.py dist_xo
```
we would have to use
```python
sugar-activity-build dist_xo
```

@quozl @walterbender should this change be included ?

reasons for proposed change - better than the alternate of having same `setup.py` in each activity.

### DIFF
--- a/bin/Makefile.am
+++ b/bin/Makefile.am
@@ -1,1 +1,1 @@
-dist_bin_SCRIPTS = sugar-activity sugar-activity-web
+dist_bin_SCRIPTS = sugar-activity sugar-activity-web sugar-activity-build
--- a/bin/sugar-activity-build
+++ b/bin/sugar-activity-build
@@ -1,0 +1,5 @@
+#!/usr/bin/env python2
+
+from sugar3.activity import bundlebuilder
+
+bundlebuilder.start()
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -622,7 +622,7 @@ def start():
 
     options = parser.parse_args()
 
-    source_dir = os.path.abspath(os.path.dirname(sys.argv[0]))
+    source_dir = os.path.abspath(os.path.dirname(os.curdir))
     config = Config(source_dir)
 
     try:


tony37:
This assumes that the user has installed a Sugar build with this 
capability. I hope that this change can be deferred until the next XO 
build is released.

Tony

Pro-Panda:
> This assumes that the user has installed a Sugar build with this capability. I hope that this change can be deferred until the next XO build is released.

Apologies, but I couldn't follow, what capability is being referred here ?

quozl:
Thanks, that's interesting.

As the old `setup.py` method would continue to work, there's no need to upgrade unless someone is working on an activity source code clone where `setup.py` has been removed.  So the key point in time is when an activity repository gets a pull request to remove `setup.py`.  This would create an incompatibility; the activity would not longer be maintainable on an older version of Sugar, and the saving of having one less file in each activity doesn't seem worth it.

I don't like the command name _sugar-activity-build_ because;
* there are many `setup.py` targets for which "build" as a verb doesn't seem to make sense; like "dist_source", "dist_xo", "genpot", and "dev",
* _sugar-activity_, _sugar-activity-web_, and _sugar-activity3_ are functionally the same, but _sugar-activity-build_ would not be.

@walterbender, what do you think?

pro-panda:
> As the old setup.py method would continue to work, there's no need to upgrade unless someone is working on an activity source code clone where setup.py has been removed. So the key point in time is when an activity repository gets a pull request to remove setup.py. This would create an incompatibility; the activity would not longer be maintainable on an older version of Sugar, and the saving of having one less file in each activity doesn't seem worth it.

Thanks, I agree on all points. But,
 - the functionality added by `setup.py` is not related to any particular activity
 - changes to `setup.py` will be too tedious to implement

> I don't like the command name sugar-activity-build because;
> - there are many setup.py targets for which "build" as a verb doesn't seem to make sense; like "dist_source", "dist_xo", "genpot", and "dev",
> - sugar-activity, sugar-activity-web, and sugar-activity3 are functionally the same, but sugar-activity-build would not be.

I agree, I'm bad at naming files. Any suggestions on an alternate name ?

quozl:
> changes to `setup.py` will be too tedious to implement

What changes?

Also, consider that any pull request to remove `setup.py` from an activity is unlikely to be accepted because it breaks the activity for development within older versions of Sugar.

tony37:
One simple solution is to put setup.py in the python path on the Sugar 
system.

Tony

quozl:
That's pretty much what Rahul has suggested in the patch, although presumably to avoid intruding on other Python packages that use `setup.py` files, Rahul has used the name `sugar-activity-build`.

walterbender:
How about sugar-activity-setup as a name?

pro-panda:
> What changes?

Reframing,
- any change, in `setup.py` (if planned to be implemented) will have to be iterated across each activity

> Also, consider that any pull request to remove setup.py from an activity is unlikely to be accepted 
> because it breaks the activity for development within older versions of Sugar.

 - Thanks, I agree. But, what is the ratio of users / organisations / schools using older versions of Sugar versus Sugar 0.112. Also, once a newer version is released, are activities developed/maintained for the previous versions as well ?

> How about sugar-activity-setup as a name?

I am in favor of this one

pro-panda:
We should close this if we don't have a concensus in favor of this change

pro-panda:
Stale. No further response from contributors in achieving a consensus. Closing

PR #398, opened 2018-07-24
Port from GObject to GLib

Pro-Panda:


quozl:
Several new E127 and E501.

Pro-Panda:
Made several PEP8 style changes and flake fixes using `autopep8`. Please re review

quozl:
Again, my preference is not to autopep8, as now the change is way larger than is practical to review without losing attentiveness to the purpose of the pull request.  It also breaks the use of flake8 comparison as a review step.

Pro-Panda:
Removed `autopep8` changes. Made flake8 fixes. 
Please re-review, thanks

Issue #399, opened 2018-08-09
Standard list of activity tags


quozl:
Recent development as part of GSoC 2018 has added *tags* to many of the core Fructose activities, with several side-effects showing up during integration;

* the List View  (<kbd>F2</kbd> <kbd>ctrl</kbd>+<kbd>2</kbd>) shows the tags in a comma separated list under the activity name,

* the List View search feature finds activities with tags,

* the tags are not translated into the configured language,

* the tags are somewhat inconsistently applied.

*tags* is briefly defined in https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/src/sugar3/bundle/__init__.py#L72

We need;

* [ ] a standard list of tags,
* [ ] translations of the tags.

CC: @vipulgupta2048 

Screenshot of List View after searching for "documen";

![screenshot_ubb_2018-08-09_15 07 06](https://user-images.githubusercontent.com/2936481/43879203-222182d2-9be6-11e8-9ac4-d05aebe9e802.png)

vipulgupta2048:
> a standard list of tags

Like @quozl mentioned and after going over a lot of activities for my project in GSoC 2018. The `tags` parameter was introduced to maintain uniformity in terms of the classification of activities. The list of tags below can be considered as standard if the community agrees on it. 

Programming, Robotics, Internet, Science, Maths, Language, Geography, Documents, Music, Media, Art, Teacher, or System.

Apart from the tags listed above, I have also proposed the use of `Tools` which can be classified as activities that are essential and increases productivity. I might not have noticed `Utilities` not being the list hence the tag being in write activity,

SO, if we could have a vote between `Utilities` or `Tools`, then we could add the same in #400

quozl:
@vipulgupta2048, actually the `tags` metadata feature was already there, it's just that your project began to use it for aslo-v3 and so you began populating it.  That's why it is starting to show up in List View.

`Tools` is as ambiguous as `Activity`.  I don't think `Tools` has any meaning.

`Utilities` would more commonly mean programs that act on system or storage components; like the Journal, Frame, Neighbourhood View, or the hidden Log and Terminal activities.

That productivity is increased is a value judgement disconnected from child learning; the Maze game will increase productivity by improving visual perception; the Implode game will increase productivity by improving problem solving ability, etc.

@walterbender, is there a software classification system used in pedagogy?

tony37:
I can see no benefit in cluttering the List view with these tags. The 
[...]

PR #400, opened 2018-08-09
Add Tools in tags category

vipulgupta2048:
Corrected sample browse activity metadata. 
Fixed address in GPLv3
Add Tools (Need suggestions on this from the community) in tags.


quozl:
As I said in https://github.com/sugarlabs/sugar-toolkit-gtk3/issues/399 my preference is not to have `Tools`, as it holds no specific meaning.

quozl:
c7a7dcea7d93d76299aa1267e55b8b5053553b8f

PR #407, opened 2019-03-16
expose mark messages

chimosky:


quozl:
Thanks.  This requires all activities to set `sugar3.speech.highlight_cb` to something other than `None` before speaking?  Could the API be more explicit?

chimosky:
> This requires all activities to set sugar3.speech.highlight_cb to something other than None before speaking? Could the API be more explicit?

Yes it does as the function responsible for text highlight is within the activity, so `sugar3.speech.highlight_cb` should be assigned to the function.

quozl:
Thanks.  In that case I can't accept it.  There are better ways to define the API.

chimosky:
> Thanks. This requires all activities to set sugar3.speech.highlight_cb to something other than None before speaking? Could the API be more explicit?

No it doesn't require all activities to set highlight_cb to a value before speaking, this only exposes mark messages which is used for text highlighting.

quozl:
Thanks.  Are you looking at the code I'm looking at?  In 973d475 the variable is assigned to `None`.  When a mark message is emitted, we will see _TypeError: 'NoneType' object is not callable_.  Only way to stop that is to require an activity to set it to something callable.

chimosky:
> Thanks. Are you looking at the code I'm looking at? In [973d475](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/973d47548af4ef5824e8a51cbdcdcf8384a26d92) the variable is assigned to `None`. When a mark message is emitted, we will see _TypeError: 'NoneType' object is not callable_. Only way to stop that is to require an activity to set it to something callable.

I agree, fixed in a859191.

quozl:
Thanks.  However, now we have multiple methods to specify callbacks; signals, and an attribute of the object.  Is there any reason you can't define a new signal and emit it?  That way there would only be one method to specify callbacks.

chimosky:
Like define a new signal after catching marked messages and connecting the signal to the desired function? 

What do you suggest?

quozl:
No, define the new signal always, as any new signals defined don't require any porting by the caller.  Call the emit in the message handler.  Signal could be "mark".

chimosky:
@quozl kindly review.

src/sugar3/speech.py
```diff
@@ -212,6 +213,7 @@ def __init__(self, **kwargs):
         self._player.connect('play', self._update_state, 'play')
         self._player.connect('stop', self._update_state, 'stop')
         self._player.connect('pause', self._update_state, 'pause')
+        self._player.connect('mark', self._marked_messages_cb)
```
quozl: Fit our code style.  Method name should be `__mark_cb`.

(`_update_state` doesn't fit our style, but no need to change it.)

src/sugar3/speech.py
```diff
@@ -220,6 +222,28 @@ def __init__(self, **kwargs):
         self._save_timeout_id = -1
         self.restore()
 
+    def _marked_messages_cb(self, player, value):
```
quozl: Move this method below `_update_state` so that overall method order is consistent with the order in which it was mentioned.

src/sugar3/speech.py
```diff
@@ -220,6 +222,28 @@ def __init__(self, **kwargs):
...
+        '''
+        This method provides access to mark message values which can be used
```
quozl: This is great documentation, but it is in the wrong place.  The method is not to be called by the activity.  The method is private to the SpeechManager class.

src/sugar3/speech.py
```diff
@@ -220,6 +222,28 @@ def __init__(self, **kwargs):
...
+        for text highlighting.(see readetexts activity)
+
+        In your Speech Toolbar class, connect the `mark` signal to your
+        own callback.
+
+        .. code-block:: python
+
+            from sugar3.speech import SpeechManager
+
+            class SpeechToolbar(Gtk.Toolbar):
```
quozl: Not all uses of SpeechManager require a SpeechToolbar, so mentioning SpeechToolbar seems overly specific.

src/sugar3/speech.py
```diff
@@ -331,7 +355,8 @@ class _GstSpeechPlayer(GObject.GObject):
     __gsignals__ = {
         'play': (GObject.SignalFlags.RUN_FIRST, None, []),
         'pause': (GObject.SignalFlags.RUN_FIRST, None, []),
-        'stop': (GObject.SignalFlags.RUN_FIRST, None, [])
+        'stop': (GObject.SignalFlags.RUN_FIRST, None, []),
+        'mark': (GObject.SignalFlags.RUN_FIRST, None, (int,))
```
quozl: Another extra comma after int?

src/sugar3/speech.py
```diff
@@ -382,6 +407,10 @@ def __pipe_message_cb(self, bus, message):
             self._pipeline = None
             power.get_power_manager().restore_suspend()
             self.emit('stop')
+        elif message.type is Gst.MessageType.ELEMENT and \
+                message.get_structure().get_name() == 'espeak-mark':
+            mark_value = message.get_structure().get_value('mark')
+            self.emit('mark', int(mark_value))
```
quozl: Why is the mark value being converted to int?

src/sugar3/speech.py
```diff
@@ -193,7 +193,8 @@ class SpeechManager(GObject.GObject):
     __gsignals__ = {
         'play': (GObject.SignalFlags.RUN_FIRST, None, []),
         'pause': (GObject.SignalFlags.RUN_FIRST, None, []),
-        'stop': (GObject.SignalFlags.RUN_FIRST, None, [])
+        'stop': (GObject.SignalFlags.RUN_FIRST, None, []),
+        'mark': (GObject.SignalFlags.RUN_FIRST, None, (int,))
```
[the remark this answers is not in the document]
chimosky: [Signal argument type usually end with a comma](https://python-gtk-3-tutorial.readthedocs.io/en/latest/objects.html#create-new-signals) because the `__gsignals__` third argument is expected to be a sequence else a `TypeError` is thrown.

src/sugar3/speech.py
chimosky: Will fix that, thanks.

src/sugar3/speech.py
chimosky: Will fix that, thanks.

src/sugar3/speech.py
chimosky: Yes the method is private to the SpeechManager class, the documentation was basically to show how to connect the signal, the necessary arguments. I can remove it.

src/sugar3/speech.py
chimosky: Yes you're right, I made an error of thinking of a particular case. I'll fix that, thanks.

src/sugar3/speech.py
chimosky: Commented on above.

src/sugar3/speech.py
chimosky: A mistake I made as the context to which I applied it to used the number of a word which it gets from the name attribute of the mark tag so this way the method being called will know which word to highlight. I'll change it and restrict that to the context in which it's being used, thanks.

src/sugar3/speech.py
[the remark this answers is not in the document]
quozl: Thanks.  Let me rephrase that.  While the upstream documentation implies we should use a `tuple` type, almost all our code in Sugar and Toolkit use a `list` type.  You can see this with `git grep SignalFlags`.  I don't mind if it all changes to `tuple` type, as long as it does it in one go, but mixing the styles is unnecessary and unpleasant.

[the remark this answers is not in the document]
chimosky: I'll make the required changes, thanks.

src/sugar3/speech.py
chimosky: @quozl waiting for your opinion here.

quozl: I can't say no to documentation, but I will say no to inconsistent documentation.  It would be great to have automatically generated documentation for this module.  But it need not happen in this pull request, because the module at the moment has almost no documentation.

chimosky:
@quozl kindly review.

chimosky:
@quozl kindly review.

quozl:
Thanks.  Merged with some changes.

You might also be interested in the other text position value in the mark message, and the other messages that gst-plugins-espeak can emit, see the source code at https://github.com/sugarlabs/gst-plugins-espeak/blob/master/src/espeak.c#L99

PR #408, opened 2019-03-18
Set bundlebuilder to honor .gitignore

pro-panda:
Partially fixes https://github.com/sugarlabs/sugar-docs/issues/91


quozl:
Sorry, but I was entirely wrong about [having to use a pristine clone](https://github.com/sugarlabs/flappy-birds-activity/issues/1#issuecomment-221725940).  The toolkit already uses the list of files tracked by git to make the bundle, and excludes other files automatically.  I've many untracked files in my clones that are still there despite me making bundles, and the bundles do not contain the files.

pro-panda:
Thanks, I agree
My test method was flawed. I had added .gitignore after some *.pyc files for already present (and tracked by git).

PR #411, opened 2019-05-15
Build for either python 2 or python 3

quozl:
`configure` and `autogen.sh` now accept `--with-python2` or `--with-python3` options, with the default of python 2.

aa8a5e7 introduced building for python 3, but broke build for python 2, requiring patching `configure.ac` prior to build.

Code was derived from rpmfusion-packager, found using a GitHub search of `configure.ac` files for both `python2` and `python3`, and is GPLv3.

Tested on Ubuntu 18.04.

Signed-off-by: James Cameron <quozl@laptop.org>

@Aniket21mathur, please test.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -15,8 +15,17 @@ GNOME_COMPILE_WARNINGS(maximum)
 
 AC_PATH_PROG([GLIB_GENMARSHAL], [glib-genmarshal])
 
-PYTHON=python3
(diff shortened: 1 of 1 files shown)


pro-panda:
Thanks, tested.

Aniket21mathur:
Thanks @pro-panda .

PR #401, opened 2018-11-19
[In Progress] Port to TelepathyGLib

tonadev:
### Explanation
Continues with #389 but it's still **in progress**. This PR ports to TelepathyGLib.

### Reason
The telepathy library does not have its bindings for Python 3, and porting Telepathy to its PyGObject binding is a pre requisite for the Port to Python 3 Project. 

*I tested it to see that activities continue working, and they do. However I encourage you to not trust me and test it by yourselves.*

PR #412, opened 2019-05-28
Port to TelepathyGLib

Aniket21mathur:
Continued from #401 (In progress pr) 
@quozl @pro-panda . Can you please test it.Thanks.

### DIFF
--- a/src/sugar3/activity/activity.py
+++ b/src/sugar3/activity/activity.py
@@ -1434,6 +1434,7 @@ class _ClientHandler(dbus.service.Object):
         dbus.service.Object.__init__(self, bus_name, path)
 
         self._prop_getters = {}
+        self._prop_setters = {}
         self._prop_getters.setdefault(CLIENT, {}).update({
             'Interfaces': lambda: list(self._interfaces),
         })
@@ -1468,6 +1469,34 @@ class _ClientHandler(dbus.service.Object):
         except Exception as e:
             logging.exception(e)
 
+    @dbus.service.method(dbus_interface=dbus.PROPERTIES_IFACE,
+                         in_signature='ss', out_signature='v')
+    def Get(self, interface_name, property_name):
+        if interface_name in self._prop_getters \
+            and property_name in self._prop_getters[interface_name]:
+                return self._prop_getters[interface_name][property_name]()
+        else:
+            logging.debug('InvalidArgument')
+
+    @dbus.service.method(dbus_interface=dbus.PROPERTIES_IFACE,
+                         in_signature='ssv', out_signature='')
+    def Set(self, interface_name, property_name, value):
+        if interface_name in self._prop_setters \
+            and property_name in self._prop_setters[interface_name]:
+                self._prop_setters[interface_name][property_name](value)
+        else:
+            logging.debug('PermissionDenied')
+
+    @dbus.service.method(dbus_interface=dbus.PROPERTIES_IFACE,
+                         in_signature='s', out_signature='a{sv}')
+    def GetAll(self, interface_name):
+        if interface_name in self._prop_getters:
+            r = {}
+            for k, v in self._prop_getters[interface_name].items():
+                r[k] = v()
+            return r
+        else:
+            logging.debug('InvalidArgument')
 
 _session = None
 
--- a/src/sugar3/activity/activity.py
+++ b/src/sugar3/activity/activity.py
@@ -217,6 +217,14 @@ N_BUS_NAME = 'org.freedesktop.Notifications'
 N_OBJ_PATH = '/org/freedesktop/Notifications'
 N_IFACE_NAME = 'org.freedesktop.Notifications'
 
+CHANNEL = TelepathyGLib.IFACE_CHANNEL
+CHANNEL_TYPE_TEXT = TelepathyGLib.IFACE_CHANNEL_TYPE_TEXT
+CLIENT = TelepathyGLib.IFACE_CLIENT
+CLIENT_HANDLER = TelepathyGLib.IFACE_CLIENT_HANDLER
+
+CONNECTION_HANDLE_TYPE_CONTACT = TelepathyGLib.HandleType.CONTACT
+CONNECTION_HANDLE_TYPE_ROOM = TelepathyGLib.HandleType.ROOM
+
 CONN_INTERFACE_ACTIVITY_PROPERTIES = 'org.laptop.Telepathy.ActivityProperties'
 
 PREVIEW_SIZE = style.zoom(300), style.zoom(225)
@@ -541,11 +549,11 @@ class Activity(Window, Gtk.Container):
         logging.debug('Activity.__got_channel_cb')
         pservice = presenceservice.get_instance()
 
-        if handle_type == TelepathyGLib.HandleType.ROOM:
+        if handle_type == CONNECTION_HANDLE_TYPE_ROOM:
             connection_name = connection_path.replace('/', '.')[1:]
             bus = dbus.SessionBus()
             channel = bus.get_object(connection_name, channel_path)
-            room_handle = channel.Get(TelepathyGLib.IFACE_CHANNEL, 'TargetHandle')
+            room_handle = channel.Get(CHANNEL, 'TargetHandle')
             mesh_instance = pservice.get_activity_by_handle(connection_path,
                                                             room_handle)
         else:
@@ -1415,36 +1423,36 @@ class Activity(Window, Gtk.Container):
 
 class _ClientHandler(dbus.service.Object):
     def __init__(self, bundle_id, got_channel_cb):
-        self._interfaces = set([TelepathyGLib.IFACE_CLIENT, TelepathyGLib.IFACE_CLIENT_HANDLER, PROPERTIES_IFACE])
+        self._interfaces = set([CLIENT, CLIENT_HANDLER, PROPERTIES_IFACE])
         self._got_channel_cb = got_channel_cb
 
         bus = dbus.Bus()
-        name = TelepathyGLib.IFACE_CLIENT + '.' + bundle_id
+        name = CLIENT + '.' + bundle_id
         bus_name = dbus.service.BusName(name, bus=bus)
 
         path = '/' + name.replace('.', '/')
         dbus.service.Object.__init__(self, bus_name, path)
 
         self._prop_getters = {}
-        self._prop_getters.setdefault(TelepathyGLib.IFACE_CLIENT, {}).update({
+        self._prop_getters.setdefault(CLIENT, {}).update({
             'Interfaces': lambda: list(self._interfaces),
         })
-        self._prop_getters.setdefault(TelepathyGLib.IFACE_CLIENT_HANDLER, {}).update({
+        self._prop_getters.setdefault(CLIENT_HANDLER, {}).update({
             'HandlerChannelFilter': self.__get_filters_cb,
         })
 
     def __get_filters_cb(self):
         logging.debug('__get_filters_cb')
         filters = {
-            TelepathyGLib.IFACE_CHANNEL + '.ChannelType': TelepathyGLib.IFACE_CHANNEL_TYPE_TEXT,
-            TelepathyGLib.IFACE_CHANNEL + '.TargetHandleType': TelepathyGLib.HandleType.CONTACT,
+            CHANNEL + '.ChannelType': CHANNEL_TYPE_TEXT,
+            CHANNEL + '.TargetHandleType': CONNECTION_HANDLE_TYPE_CONTACT,
         }
         filter_dict = dbus.Dictionary(filters, signature='sv')
         logging.debug('__get_filters_cb %r' % dbus.Array([filter_dict],
                                                          signature='a{sv}'))
         return dbus.Array([filter_dict], signature='a{sv}')
 
-    @dbus.service.method(dbus_interface=TelepathyGLib.IFACE_CLIENT_HANDLER,
+    @dbus.service.method(dbus_interface=CLIENT_HANDLER,
                          in_signature='ooa(oa{sv})aota{sv}', out_signature='')
     def HandleChannels(self, account, connection, channels, requests_satisfied,
                        user_action_time, handler_info):
@@ -1453,9 +1461,9 @@ class _ClientHandler(dbus.service.Object):
                           user_action_time, handler_info))
         try:
             for object_path, properties in channels:
-                channel_type = properties[TelepathyGLib.IFACE_CHANNEL + '.ChannelType']
-                handle_type = properties[TelepathyGLib.IFACE_CHANNEL + '.TargetHandleType']
-                if channel_type == TelepathyGLib.IFACE_CHANNEL_TYPE_TEXT:
+                channel_type = properties[CHANNEL + '.ChannelType']
+                handle_type = properties[CHANNEL + '.TargetHandleType']
+                if channel_type == CHANNEL_TYPE_TEXT:
                     self._got_channel_cb(connection, object_path, handle_type)
         except Exception as e:
             logging.exception(e)
--- a/src/sugar3/activity/activity.py
+++ b/src/sugar3/activity/activity.py
@@ -183,7 +183,6 @@ from gi.repository import TelepathyGLib
 import dbus
 import dbus.service
 from dbus import PROPERTIES_IFACE
-from telepathy.server import DBusProperties
 
 from sugar3 import util
 from sugar3 import power
@@ -1414,7 +1413,7 @@ class Activity(Window, Gtk.Container):
         Gdk.flush()
 
 
-class _ClientHandler(dbus.service.Object, DBusProperties):
+class _ClientHandler(dbus.service.Object):
     def __init__(self, bundle_id, got_channel_cb):
         self._interfaces = set([TelepathyGLib.IFACE_CLIENT, TelepathyGLib.IFACE_CLIENT_HANDLER, PROPERTIES_IFACE])
         self._got_channel_cb = got_channel_cb
@@ -1425,12 +1424,12 @@ class _ClientHandler(dbus.service.Object, DBusProperties):
 
         path = '/' + name.replace('.', '/')
         dbus.service.Object.__init__(self, bus_name, path)
-        DBusProperties.__init__(self)
 
-        self._implement_property_get(TelepathyGLib.IFACE_CLIENT, {
+        self._prop_getters = {}
+        self._prop_getters.setdefault(TelepathyGLib.IFACE_CLIENT, {}).update({
             'Interfaces': lambda: list(self._interfaces),
         })
-        self._implement_property_get(TelepathyGLib.IFACE_CLIENT_HANDLER, {
+        self._prop_getters.setdefault(TelepathyGLib.IFACE_CLIENT_HANDLER, {}).update({
             'HandlerChannelFilter': self.__get_filters_cb,
         })
 
--- a/src/sugar3/activity/activity.py
+++ b/src/sugar3/activity/activity.py
@@ -172,22 +172,18 @@ import json
 import gi
 gi.require_version('Gtk', '3.0')
 gi.require_version('Gdk', '3.0')
+gi.require_version('TelepathyGLib', '0.12')
 gi.require_version('SugarExt', '1.0')
 
 from gi.repository import GLib
 from gi.repository import GObject
 from gi.repository import Gdk
 from gi.repository import Gtk
+from gi.repository import TelepathyGLib
 import dbus
 import dbus.service
 from dbus import PROPERTIES_IFACE
 from telepathy.server import DBusProperties
-from telepathy.interfaces import CHANNEL, \
-    CHANNEL_TYPE_TEXT, \
-    CLIENT, \
-    CLIENT_HANDLER
-from telepathy.constants import CONNECTION_HANDLE_TYPE_CONTACT
-from telepathy.constants import CONNECTION_HANDLE_TYPE_ROOM
 
 from sugar3 import util
 from sugar3 import power
@@ -546,11 +542,11 @@ class Activity(Window, Gtk.Container):
         logging.debug('Activity.__got_channel_cb')
         pservice = presenceservice.get_instance()
 
-        if handle_type == CONNECTION_HANDLE_TYPE_ROOM:
+        if handle_type == TelepathyGLib.HandleType.ROOM:
             connection_name = connection_path.replace('/', '.')[1:]
             bus = dbus.SessionBus()
             channel = bus.get_object(connection_name, channel_path)
-            room_handle = channel.Get(CHANNEL, 'TargetHandle')
+            room_handle = channel.Get(TelepathyGLib.IFACE_CHANNEL, 'TargetHandle')
             mesh_instance = pservice.get_activity_by_handle(connection_path,
                                                             room_handle)
         else:
@@ -1420,36 +1416,36 @@ class Activity(Window, Gtk.Container):
 
 class _ClientHandler(dbus.service.Object, DBusProperties):
     def __init__(self, bundle_id, got_channel_cb):
-        self._interfaces = set([CLIENT, CLIENT_HANDLER, PROPERTIES_IFACE])
+        self._interfaces = set([TelepathyGLib.IFACE_CLIENT, TelepathyGLib.IFACE_CLIENT_HANDLER, PROPERTIES_IFACE])
         self._got_channel_cb = got_channel_cb
 
         bus = dbus.Bus()
-        name = CLIENT + '.' + bundle_id
+        name = TelepathyGLib.IFACE_CLIENT + '.' + bundle_id
         bus_name = dbus.service.BusName(name, bus=bus)
 
         path = '/' + name.replace('.', '/')
         dbus.service.Object.__init__(self, bus_name, path)
         DBusProperties.__init__(self)
 
-        self._implement_property_get(CLIENT, {
+        self._implement_property_get(TelepathyGLib.IFACE_CLIENT, {
             'Interfaces': lambda: list(self._interfaces),
         })
-        self._implement_property_get(CLIENT_HANDLER, {
+        self._implement_property_get(TelepathyGLib.IFACE_CLIENT_HANDLER, {
             'HandlerChannelFilter': self.__get_filters_cb,
         })
 
     def __get_filters_cb(self):
         logging.debug('__get_filters_cb')
         filters = {
-            CHANNEL + '.ChannelType': CHANNEL_TYPE_TEXT,
-            CHANNEL + '.TargetHandleType': CONNECTION_HANDLE_TYPE_CONTACT,
+            TelepathyGLib.IFACE_CHANNEL + '.ChannelType': TelepathyGLib.IFACE_CHANNEL_TYPE_TEXT,
+            TelepathyGLib.IFACE_CHANNEL + '.TargetHandleType': TelepathyGLib.HandleType.CONTACT,
         }
         filter_dict = dbus.Dictionary(filters, signature='sv')
         logging.debug('__get_filters_cb %r' % dbus.Array([filter_dict],
                                                          signature='a{sv}'))
         return dbus.Array([filter_dict], signature='a{sv}')
 
-    @dbus.service.method(dbus_interface=CLIENT_HANDLER,
+    @dbus.service.method(dbus_interface=TelepathyGLib.IFACE_CLIENT_HANDLER,
                          in_signature='ooa(oa{sv})aota{sv}', out_signature='')
     def HandleChannels(self, account, connection, channels, requests_satisfied,
                        user_action_time, handler_info):
@@ -1458,9 +1454,9 @@ class _ClientHandler(dbus.service.Object, DBusProperties):
                           user_action_time, handler_info))
         try:
             for object_path, properties in channels:
-                channel_type = properties[CHANNEL + '.ChannelType']
-                handle_type = properties[CHANNEL + '.TargetHandleType']
-                if channel_type == CHANNEL_TYPE_TEXT:
+                channel_type = properties[TelepathyGLib.IFACE_CHANNEL + '.ChannelType']
+                handle_type = properties[TelepathyGLib.IFACE_CHANNEL + '.TargetHandleType']
+                if channel_type == TelepathyGLib.IFACE_CHANNEL_TYPE_TEXT:
                     self._got_channel_cb(connection, object_path, handle_type)
         except Exception as e:
             logging.exception(e)
--- a/src/sugar3/presence/activity.py
+++ b/src/sugar3/presence/activity.py
@@ -43,7 +43,6 @@ CHANNEL_TYPE_TEXT = TelepathyGLib.IFACE_CHANNEL_TYPE_TEXT
 CONNECTION = TelepathyGLib.IFACE_CONNECTION
 PROPERTIES_INTERFACE = TelepathyGLib.IFACE_PROPERTIES_INTERFACE
 
-# from telepathy.constants import
 CHANNEL_GROUP_FLAG_CHANNEL_SPECIFIC_HANDLES = \
     TelepathyGLib.ChannelGroupFlags.CHANNEL_SPECIFIC_HANDLES
 HANDLE_TYPE_CONTACT = TelepathyGLib.HandleType.CONTACT
@@ -107,6 +106,8 @@ class Activity(GObject.GObject):
 
         self._account_path = account_path
         self.telepathy_conn = connection
+        self.telepathy_text_chan = None
+        self.telepathy_tubes_chan = None
 
         self.room_handle = room_handle
         self._join_command = None
@@ -278,8 +279,7 @@ class Activity(GObject.GObject):
         if not self._joined:
             raise RuntimeError('Cannot invite a buddy to an activity that is'
                                'not shared.')
-        chan = dbus.Interface(self.obj, CHANNEL)
-        chan.AddMembers(
+        self.telepathy_text_chan[CHANNEL].AddMembers(
             [buddy.contact_handle], message,
             dbus_interface=CHANNEL_INTERFACE_GROUP,
             reply_handler=partial(
@@ -296,14 +296,15 @@ class Activity(GObject.GObject):
         _logger.debug('%r: Join finished %r' % (self, error))
         if error is not None:
             self.emit('joined', error is None, str(error))
-        self.obj = join_command.obj
+        self.telepathy_text_chan = join_command.text_channel
+        self.telepathy_tubes_chan = join_command.tubes_channel
         self._channel_self_handle = join_command.channel_self_handle
         self._text_channel_group_flags = join_command.text_channel_group_flags
         self._start_tracking_buddies()
         self._start_tracking_channel()
 
     def _start_tracking_buddies(self):
-        group = dbus.Interface(self.obj, CHANNEL_INTERFACE_GROUP)
+        group = self.telepathy_text_chan[CHANNEL_INTERFACE_GROUP]
 
         group.GetAllMembers(reply_handler=self.__get_all_members_cb,
                             error_handler=self.__error_handler_cb)
@@ -312,7 +313,7 @@ class Activity(GObject.GObject):
                                 self.__text_channel_members_changed_cb)
 
     def _start_tracking_channel(self):
-        channel = dbus.Interface(self.obj, CHANNEL)
+        channel = self.telepathy_text_chan[CHANNEL]
         channel.connect_to_signal('Closed', self.__text_channel_closed_cb)
 
     def __get_all_members_cb(self, members, local_pending, remote_pending):
@@ -338,7 +339,7 @@ class Activity(GObject.GObject):
         if self._text_channel_group_flags & \
                 CHANNEL_GROUP_FLAG_CHANNEL_SPECIFIC_HANDLES:
 
-            group = dbus.Interface(self.obj, CHANNEL_INTERFACE_GROUP)
+            group = self.telepathy_text_chan[CHANNEL_INTERFACE_GROUP]
             group.GetHandleOwners(input_handles,
                                   reply_handler=get_handle_owners_cb,
                                   error_handler=self.__error_handler_cb)
@@ -428,8 +429,8 @@ class Activity(GObject.GObject):
         if error is None:
             self._joined = True
             self.room_handle = share_command.room_handle
-            self.obj = share_command.obj
-            self.channel_path = share_command.channel_path
+            self.telepathy_text_chan = share_command.text_channel
+            self.telepathy_tubes_chan = share_command.tubes_channel
             self._channel_self_handle = share_command.channel_self_handle
             self._text_channel_group_flags = \
                 share_command.text_channel_group_flags
@@ -478,8 +479,9 @@ class Activity(GObject.GObject):
         """
         bus_name = self.telepathy_conn.requested_bus_name
         connection_path = self.telepathy_conn.object_path
-        channels = [self.channel_path,
-                    self.channel_path]
+        channels = [self.telepathy_text_chan.object_path,
+                    self.telepathy_tubes_chan.object_path]
+
         _logger.debug('%r: bus name is %s, connection is %s, channels are %r' %
                       (self, bus_name, connection_path, channels))
         return bus_name, connection_path, channels
@@ -492,8 +494,7 @@ class Activity(GObject.GObject):
     def leave(self):
         """Leave this shared activity"""
         _logger.debug('%r: leaving' % self)
-        chan = dbus.Interface(self.obj, CHANNEL)
-        chan.Close()
+        self.telepathy_text_chan[CHANNEL].Close()
 
 
 class _BaseCommand(GObject.GObject):
@@ -505,7 +506,9 @@ class _BaseCommand(GObject.GObject):
     def __init__(self):
         GObject.GObject.__init__(self)
 
+        self.text_channel = None
         self.text_channel_group_flags = None
+        self.tubes_channel = None
         self.room_handle = None
         self.channel_self_handle = None
 
@@ -544,9 +547,10 @@ class _ShareCommand(_BaseCommand):
             self._finished = True
             self.emit('finished', error)
             return
-        self.obj = join_command.obj
+
+        self.text_channel = join_command.text_channel
         self.text_channel_group_flags = join_command.text_channel_group_flags
-        self.channel_path = join_command.channel_path
+        self.tubes_channel = join_command.tubes_channel
         self.channel_self_handle = join_command.channel_self_handle
 
         self._connection.AddActivity(
@@ -585,8 +589,10 @@ class _JoinCommand(_BaseCommand):
                              dbus_interface=PROPERTIES_IFACE)
 
     def __get_self_handle_cb(self, handle):
+        _logger.error('_JoinCommand.__get_self_handle_cb')
         self._global_self_handle = handle
 
+        self._text_ready = False
         self._connection.RequestChannel(
             CHANNEL_TYPE_TEXT,
             HANDLE_TYPE_ROOM,
@@ -595,24 +601,81 @@ class _JoinCommand(_BaseCommand):
             error_handler=self.__error_handler_cb,
             dbus_interface=CONNECTION)
 
-    def __create_text_channel_cb(self, channel_path):
-        self.obj = dbus.Bus().get_object(self._connection.requested_bus_name, channel_path)
-        self.channel_path = channel_path
-        self._valid_interfaces = set()
-        channel_type = dbus.Interface(self.obj, CHANNEL).GetChannelType()
-        self._valid_interfaces.add(channel_type)
+        self._tubes_ready = False
+        self._connection.RequestChannel(
+            CHANNEL_TYPE_TUBES,
+            HANDLE_TYPE_ROOM,
+            self.room_handle,
+            True,
+            reply_handler=self.__create_tubes_channel_cb,
+            error_handler=self.__tubes_error_handler_cb,
+            dbus_interface=CONNECTION)
 
-        dbus.Interface(self.obj, CHANNEL).GetInterfaces(
-            reply_handler=self.__text_channel_ready_cb,
+    def __create_text_channel_cb(self, channel_path):
+        _logger.error('_JoinCommand.__create_text_channel_cb %s' % channel_path)
+        self.text_channel = {}
+        self.text_proxy = dbus.Bus().get_object(
+            self._connection.requested_bus_name, channel_path)
+        self.text_channel[PROPERTIES_IFACE] = dbus.Interface(
+            self.text_proxy, PROPERTIES_IFACE)
+        self.text_channel[CHANNEL_TYPE_TEXT] = dbus.Interface(self.text_proxy, CHANNEL_TYPE_TEXT)
+        self.text_channel[CHANNEL] = dbus.Interface(self.text_proxy, CHANNEL)
+        self.text_channel[CHANNEL].GetInterfaces(
+            reply_handler=self.__text_get_interfaces_reply_cb,
             error_handler=self.__error_handler_cb)
 
-    def __text_channel_ready_cb(self, interfaces):
-        self._valid_interfaces.update(interfaces)
-        self._add_self_to_channel()
+    def __text_get_interfaces_reply_cb(self, interfaces):
+        _logger.error('_JoinCommand.__text_get_interfaces_reply_cb %r' % interfaces)
+        for interface in interfaces:
+            self.text_channel[interface] = dbus.Interface(
+                self.text_proxy, interface)
+        _logger.debug('%r: Text channel is ready' % (self))
+        self._text_ready = True
+        self._ready()
+
+    def __create_tubes_channel_cb(self, channel_path):
+        _logger.error('_JoinCommand.__create_tubes_channel_cb %s' % channel_path)
+        self.tubes_channel = {}
+        self.tubes_proxy = dbus.Bus().get_object(
+            self._connection.requested_bus_name, channel_path)
+        self.tubes_channel[PROPERTIES_IFACE] = dbus.Interface(
+            self.tubes_proxy, PROPERTIES_IFACE)
+        self.tubes_channel[CHANNEL_TYPE_TUBES] = dbus.Interface(self.tubes_proxy, CHANNEL_TYPE_TUBES)
+        self.tubes_channel[CHANNEL] = dbus.Interface(self.tubes_proxy, CHANNEL)
+        self.tubes_channel[CHANNEL].GetInterfaces(
+            reply_handler=self.__tubes_get_interfaces_reply_cb,
+            error_handler=self.__error_handler_cb)
+
+    def __tubes_get_interfaces_reply_cb(self, interfaces):
+        _logger.error('_JoinCommand.__tubes_get_interfaces_reply_cb %r' % interfaces)
+        for interface in interfaces:
+            self.tubes_channel[interface] = dbus.Interface(
+                self.tubes_proxy, interface)
+        _logger.debug('%r: Tubes channel is ready' % (self))
+        self._tubes_ready = True
+        self._ready()
+
+    def __tubes_error_handler_cb(self, error):
+        if (error.get_dbus_name() ==
+                'org.freedesktop.Telepathy.Error.NotImplemented'):
+            self._tubes_supported = False
+            self._ready()
+        else:
+            self._finished = True
+            self.emit('finished', error)
 
     def __error_handler_cb(self, error):
         self._finished = True
         self.emit('finished', error)
+
+    def _ready(self):
+        if not self._text_ready or \
+                (self._tubes_supported and not self._tubes_ready):
+            return
+
+        _logger.error('%r: finished setting up channel' % self)
+
+        self._add_self_to_channel()
 
     def __text_channel_group_flags_changed_cb(self, added, removed):
         _logger.debug(
@@ -625,7 +688,7 @@ class _JoinCommand(_BaseCommand):
         # FIXME: cope with non-Group channels here if we want to support
         # non-OLPC-compatible IMs
 
-        group = dbus.Interface(self.obj, CHANNEL_INTERFACE_GROUP)
+        group = self.text_channel[CHANNEL_INTERFACE_GROUP]
 
         def got_all_members(members, local_pending, remote_pending):
             _logger.debug('got_all_members members %r local_pending %r '
@@ -689,10 +752,10 @@ class _JoinCommand(_BaseCommand):
 
         # Use RoomConfig1 to configure the text channel. If this
         # doesn't exist, fall-back on old-style PROPERTIES_INTERFACE.
-        if CONN_INTERFACE_ROOM_CONFIG in self._valid_interfaces:
+        if CONN_INTERFACE_ROOM_CONFIG in self.text_channel:
             self.__update_room_config()
-        elif PROPERTIES_INTERFACE in self._valid_interfaces:
-            dbus.Interface(self.obj, PROPERTIES_INTERFACE).ListProperties(
+        elif PROPERTIES_INTERFACE in self.text_channel:
+            self.text_channel[PROPERTIES_INTERFACE].ListProperties(
                 reply_handler=self.__list_properties_cb,
                 error_handler=self.__error_handler_cb)
         else:
@@ -717,7 +780,7 @@ class _JoinCommand(_BaseCommand):
             # don't appear in server room lists
             'Private': True,
         }
-        room_cfg = dbus.Interface(self.obj, CONN_INTERFACE_ROOM_CONFIG)
+        room_cfg = self.text_channel[CONN_INTERFACE_ROOM_CONFIG]
         room_cfg.UpdateConfiguration(props,
                                      reply_handler=self.__room_cfg_updated_cb,
                                      error_handler=self.__room_cfg_error_cb)
@@ -763,7 +826,7 @@ class _JoinCommand(_BaseCommand):
         # supported here - raise an error?
 
         if props_to_set:
-            dbus.Interface(self.obj, PROPERTIES_INTERFACE).SetProperties(
+            self.text_channel[PROPERTIES_INTERFACE].SetProperties(
                 props_to_set, reply_handler=self.__set_properties_cb,
                 error_handler=self.__error_handler_cb)
         else:
--- a/src/sugar3/presence/activity.py
+++ b/src/sugar3/presence/activity.py
@@ -28,24 +28,12 @@ from functools import partial
 import dbus
 from dbus import PROPERTIES_IFACE
 from gi.repository import GObject
-from telepathy.client import Channel
-from telepathy.interfaces import CHANNEL, \
-    CHANNEL_INTERFACE_GROUP, \
-    CHANNEL_TYPE_TUBES, \
-    CHANNEL_TYPE_TEXT, \
-    CONNECTION, \
-    PROPERTIES_INTERFACE
-from telepathy.constants import CHANNEL_GROUP_FLAG_CHANNEL_SPECIFIC_HANDLES, \
-    HANDLE_TYPE_ROOM, \
-    HANDLE_TYPE_CONTACT, \
-    PROPERTY_FLAG_WRITE
+from gi.repository import TelepathyGLib
 
 from sugar3.presence.buddy import Buddy
 
 CONN_INTERFACE_ACTIVITY_PROPERTIES = 'org.laptop.Telepathy.ActivityProperties'
 CONN_INTERFACE_BUDDY_INFO = 'org.laptop.Telepathy.BuddyInfo'
-CONN_INTERFACE_ROOM_CONFIG = \
-    'org.freedesktop.Telepathy.Channel.Interface.RoomConfig1'
 
 _logger = logging.getLogger('sugar3.presence.activity')
 
@@ -254,7 +242,7 @@ class Activity(GObject.GObject):
         return list(self._joined_buddies.values())
 
     def get_buddy_by_handle(self, handle):
-        """Retrieve the Buddy object given a telepathy handle.
+        """Retrieve the Buddy object given a TelepathyGLib handle.
 
         buddy object paths are cached in self._handle_to_buddy_path,
         so we can get the buddy without calling PS.
@@ -276,7 +264,7 @@ class Activity(GObject.GObject):
                                'not shared.')
         self.telepathy_text_chan.AddMembers(
             [buddy.contact_handle], message,
-            dbus_interface=CHANNEL_INTERFACE_GROUP,
+            dbus_interface=TelepathyGLib.IFACE_CHANNEL_INTERFACE_GROUP,
             reply_handler=partial(
                 self.__invite_cb, response_cb),
             error_handler=partial(self.__invite_cb, response_cb))
@@ -299,7 +287,7 @@ class Activity(GObject.GObject):
         self._start_tracking_channel()
 
     def _start_tracking_buddies(self):
-        group = self.telepathy_text_chan[CHANNEL_INTERFACE_GROUP]
+        group = self.telepathy_text_chan[TelepathyGLib.IFACE_CHANNEL_INTERFACE_GROUP]
 
         group.GetAllMembers(reply_handler=self.__get_all_members_cb,
                             error_handler=self.__error_handler_cb)
@@ -308,7 +296,7 @@ class Activity(GObject.GObject):
                                 self.__text_channel_members_changed_cb)
 
     def _start_tracking_channel(self):
-        channel = self.telepathy_text_chan[CHANNEL]
+        channel = self.telepathy_text_chan[TelepathyGLib.IFACE_CHANNEL]
         channel.connect_to_signal('Closed', self.__text_channel_closed_cb)
 
     def __get_all_members_cb(self, members, local_pending, remote_pending):
@@ -326,15 +314,15 @@ class Activity(GObject.GObject):
     def _resolve_handles(self, input_handles, reply_cb):
         def get_handle_owners_cb(handles):
             self.telepathy_conn.InspectHandles(
-                HANDLE_TYPE_CONTACT, handles,
+                TelepathyGLib.HandleType.CONTACT, handles,
                 reply_handler=reply_cb,
                 error_handler=self.__error_handler_cb,
-                dbus_interface=CONNECTION)
+                dbus_interface=TelepathyGLib.IFACE_CONNECTION)
 
         if self._text_channel_group_flags & \
-                CHANNEL_GROUP_FLAG_CHANNEL_SPECIFIC_HANDLES:
+                TelepathyGLib.ChannelGroupFlags.CHANNEL_SPECIFIC_HANDLES:
 
-            group = self.telepathy_text_chan[CHANNEL_INTERFACE_GROUP]
+            group = self.telepathy_text_chan[TelepathyGLib.IFACE_CHANNEL_INTERFACE_GROUP]
             group.GetHandleOwners(input_handles,
                                   reply_handler=get_handle_owners_cb,
                                   error_handler=self.__error_handler_cb)
@@ -522,11 +510,11 @@ class _ShareCommand(_BaseCommand):
 
     def run(self):
         self._connection.RequestHandles(
-            HANDLE_TYPE_ROOM,
+            TelepathyGLib.HandleType.ROOM,
             [self._activity_id],
             reply_handler=self.__got_handles_cb,
             error_handler=self.__error_handler_cb,
-            dbus_interface=CONNECTION)
+            dbus_interface=TelepathyGLib.IFACE_CONNECTION)
 
     def __got_handles_cb(self, handles):
         logging.debug('__got_handles_cb %r' % handles)
@@ -578,7 +566,7 @@ class _JoinCommand(_BaseCommand):
         if self._finished:
             raise RuntimeError('This command has already finished')
 
-        self._connection.Get(CONNECTION, 'SelfHandle',
+        self._connection.Get(TelepathyGLib.IFACE_CONNECTION, 'SelfHandle',
                              reply_handler=self.__get_self_handle_cb,
                              error_handler=self.__error_handler_cb,
                              dbus_interface=PROPERTIES_IFACE)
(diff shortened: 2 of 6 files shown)


quozl:
I'm puzzled by 5cc44e2, and don't know the code here very well; can you explain it further?

Aniket21mathur:
> I'm puzzled by [5cc44e2](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/5cc44e204ebc74ce10c1e1e7b30f1bd1fb547ff1), and don't know the code here very well; can you explain it further?

Yeah, sure.
As suggested by you I had a look at the sources.
The `_ClientHandler` class uses inheritance of `DBusProperties`.
```
class _ClientHandler(dbus.service.Object, DBusProperties):
    def __init__(self, bundle_id, got_channel_cb):
        self._interfaces = set([CLIENT, CLIENT_HANDLER, PROPERTIES_IFACE])
        self._got_channel_cb = got_channel_cb
```
which is used to call  `_implement_property_get` method.
```
DBusProperties.__init__(self)

        self._implement_property_get(CLIENT, {
            'Interfaces': lambda: list(self._interfaces),
        })
        self._implement_property_get(CLIENT_HANDLER, {
            'HandlerChannelFilter': self.__get_filters_cb,
        })
```
`DBusProperties.__init__(self)` calls out for the constructor method of the DbusProperties. Which from the teleapthy source code is 
```
class DBusProperties(dbus.service.Interface):
    def __init__(self):
        if not getattr(self, '_interfaces', None):
            self._interfaces = set()

        self._interfaces.add(dbus.PROPERTIES_IFACE)

        if not getattr(self, '_immutable_properties', None):
            self._immutable_properties = dict()

        if not getattr(self, '_prop_getters', None):
            self._prop_getters = {}
            self._prop_setters = {}
```
After calling of this method no change happens to the `self._interfaces` attribute as it is already
defined in the `_ClientHandler` class.

```
        self._interfaces = set([CLIENT, CLIENT_HANDLER, PROPERTIES_IFACE])
        self._got_channel_cb = got_channel_cb
```
self._immutable_properties, self._prop_getters and self._prop_setters attributes are declared.

Proceeding further 
```
        self._implement_property_get(TelepathyGLib.IFACE_CLIENT, {
            'Interfaces': lambda: list(self._interfaces),
        })
        self._implement_property_get(TelepathyGLib.IFACE_CLIENT_HANDLER, {
            'HandlerChannelFilter': self.__get_filters_cb,
        })
```
These call the `_implement_property_get` method of DBusProperties class.
```
 def _implement_property_get(self, iface, dict):
        self._prop_getters.setdefault(iface, {}).update(dict)
```
This method simply adds a key-value pair to the _prop_getters attribute.
With `TelepathyGLib.IFACE_CLIENT` as iface and `{
            'Interfaces': lambda: list(self._interfaces),
        }' as dict for the first call and same goes for second call.

**No other methods of DBusProperties are used in the _ClientHandler class **
So I changed the code according defined the `_prop_getters` attribute by hand and set its value just like `_interfaces`.

```
self._prop_getters = {}
        self._prop_getters.setdefault(TelepathyGLib.IFACE_CLIENT, {}).update({
            'Interfaces': lambda: list(self._interfaces),
        })
        self._prop_getters.setdefault(TelepathyGLib.IFACE_CLIENT_HANDLER, {}).update({
            'HandlerChannelFilter': self.__get_filters_cb,
        })
```
just like 

```
        self._interfaces = set([TelepathyGLib.IFACE_CLIENT, TelepathyGLib.IFACE_CLIENT_HANDLER, PROPERTIES_IFACE])
        self._got_channel_cb = got_channel_cb
```
Thanks!

quozl:
Thanks.  I agree so far.

Could I put into words the context of the execution? Looking at;
https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/src/sugar3/activity/activity.py#L425
perhaps this code is executed when the activity is started by a local user in response to an invitation from a remote user, and the purpose of the `_ClientHandler` is to receive the _HandleChannels_ remote procedure call, and pass the arguments in to `__got_channel_cb` before the activity starts, i.e.  before the Activity instance returns control to the caller.  Using a nested main loop.

Have you tested activities started by invitation and checked logs?

Aniket21mathur:
Thanks. Agreed.
Yes I tried testing it using the Chat activity and rdesktop.
I got the option to invite in the neighbourhood view of one instance, but received no invitation in the other instance. (Before I used to get a blinking icon on the top left coner.)
I use the log-activity to check for logs. I got the following

chimosky:
Tested against master branch and not port branch, activity doesn't show in the neighbourhood when shared. 64f58d39926f handles the case of `org.freedesktop.DBus.Error.UnknownMethod`.

PS: I'm yet to setup the python3 branch. From your branch, the instance being tested against isn't seen in the neighbourhood, but the instance being tested with is shown in the neighbourhood of the instance being tested against.

Aniket21mathur:
@chimosky Can you please share your logs?

chimosky:
Nothing shown in log.

quozl:
Comparing the `dbus-monitor` output would be useful, as it would show the difference between the working and non-working configuration, as far as the Telepathy API is concerned.

Aniket21mathur:
@quozl @chimosky doesn't 3725c302d66475f0f39a899522e3823da3b72b55 again makes a dependency on telepathy?

quozl:
@Aniket21mathur, yes, it does.  My understanding was that this was to verify the cause of the traceback that was observed in testing.  It isn't acceptable as a final solution.

Aniket21mathur:
Thanks. I am also getting this in shell.logs
```
dbus.exceptions.DBusException: org.freedesktop.DBus.Python.ValueError: Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/dbus/service.py", line 707, in _message_cb
    retval = candidate_method(self, *args, **keywords)
  File "/usr/local/lib/python2.7/dist-packages/sugar3/activity/activityservice.py", line 71, in InviteContact
    self._activity.invite(account_path, contact_id)
  File "/usr/local/lib/python2.7/dist-packages/sugar3/activity/activity.py", line 1093, in invite
    self.share(True)
  File "/usr/local/lib/python2.7/dist-packages/sugar3/activity/activity.py", line 1117, in share
    pservice.share_activity(self, private=private)
  File "/usr/local/lib/python2.7/dist-packages/sugar3/presence/presenceservice.py", line 200, in share_activity
    activity.get_id())
ValueError: Activity ee713f7baa1bf38601beb0f5961b3a60a98aa845 is already tracked
```

quozl:
In the shell.log of which instance; the inviter, or the responder?

Aniket21mathur:
> In the shell.log of which instance; the inviter, or the responder?

The inviter.

quozl:
I've no specific advice; I'll speak in general.  This is a very complex multi-process environment.  It isn't practical to debug by looking at logs in the absence of other data.  I suggest looking at the remote procedure calls in `dbus-monitor` to correlate against the logs.  As it is shell.log you quoted from; this is the Sugar shell _jarabe_, which maintains the neighbourhood view.  Jarabe will have received D-Bus traffic from Telepathy in response to login of the other user, and the activity you've run.  The debugging environment variables should also help to increase the information in the logs.  Your logs from 20 hours ago are also of the shell.  You'll probably need to correlate all sources of information; activity debugging logs, shell debugging logs, and dbus-monitor on both virtual machines, while performing only one user action, such as sharing the activity, or an invite.  You must also restart Sugar after changing the toolkit.  An alternative is to use a local copy of modified toolkit inside the activity directory, and keep the system toolkit unchanged so that Jarabe has a known-good toolkit to use.  That way you would only be testing the activity and toolkit changes.

quozl:
Some useful Telepathy debugging links; best was Telepathy project, worst was Sugar Labs;

* https://telepathy.freedesktop.org/resources/debugging/
* https://telepathy.freedesktop.org/wiki/DeveloperNotes/Debugging/
* https://wiki.gnome.org/action/show/Apps/Empathy/Debugging
* https://community.kde.org/KTp/FAQ#Providing_debug
* https://community.kde.org/KTp/Troubleshooting
* https://wiki.sugarlabs.org/go/BugSquad/Telepathy_Debugging
* https://wiki.sugarlabs.org/go/BugSquad/Get_Logs

In the tests you are doing with adjacent virtual machines, you are using Salut as the connection manager.

quozl:
I've been able to capture Salut debugging on two adjacent Debian VMs.  Method was;

* set up remote access over SSH or mosh,
* edit `/home/guest/.sugar/default/debug` and add these lines to the end;
```
G_MESSAGES_DEBUG=all SALUT_DEBUG=all \
    /usr/lib/telepathy/telepathy-salut 2>&1 >/tmp/salut.$(date +%s).log &
```
* log out of Sugar,
* using remote access, wait for _telepathy-salut_ to exit or stop it with `kill`,
* log in.

Actions by Salut are then logged to the file in `/tmp`.  Sharing an activity is noisy.  Hope that helps!

Aniket21mathur:
Thanks. :-)

Aniket21mathur:
In my process of testing, I figured out some things -
~ The alternative that I suggested for DBusProperties as 5cc44e204ebc74ce10c1e1e7b30f1bd1fb547ff1
is not working. It might be possible that I missed something, I will look at the sources again.
~ There is something broken or not correct in the port of src/sugar3/presence/activity.py, as keeping it's bindings to telepathy and porting the other files to TelepathyGlib (excluding some non-replacable calls like DbusProperties) results in collaboration working fine. I tested it with Chat and Maze activity, Ubuntu 18.04 using two instances.

quozl:
> something broken or not correct in the port of `src/sugar3/presence/activity.py`, as keeping it's bindings to _telepathy_ and porting the other files to _TelepathyGlib_ (excluding some non-replacable calls like _DbusProperties_) results in collaboration working fine

That's useful data, thanks.  Possible causes to consider are;
* fault introduced in porting,
* latent fault in our code that was ignored by the static binding, and exposed by the introspection binding,
* fault in the introspection binding package,
* our failure to port from the static binding for dbus to the introspection binding DBus or DBusGLib,
* our failure to find a replacement for _DbusProperties_ semantics; which is probably critical for any messages to be received by the handler.

I suggest;
* redefine the immediate problem from "collaboration does not work" to something more precise like "shared activity is not shown in neighbourhood view of peers" or "no reply is received from the _foo_ call",
* verify if messages are received by the handlers,
* test between a vm with the changes, and a vm without the changes, to identify which changes affect which roles,
* print and compare the constants in `presence/activity.py` used by the static binding and the introspection binding, (I've just done this, no errors),
* add temporary instrumentation to `presence/activity.py` to track what is happening (e.g. `logger.error` or `print >>sys.stderr`),
* run that instrumentation with the static binding, and the introspection binding, and compare the logs,

Also, I note that;
* `TelepathyGLib.IFACE_DBUS_PROPERTIES` and `dbus.PROPERTIES_IFACE` refer to the same bus name, and the code varies as to which is used,

Aniket21mathur:
Thanks. Agreed.
Another test approach that I prefer is to simply check line by line changes, that would tell us changing which part of code is causing problem.

quozl:
Yes, I had that in my draft before sending my comment, but after looking through `presence/activity.py` found that most of the changes were to the constants, and the constants ended up in my analysis as being equal.  Hack follows;

```python
from gi.repository import TelepathyGLib
import telepathy

def compare(a, b):
    exec('av = %s' % a)
    exec('bv = %s' % a)
    if av == bv:
        print '\033[7mpass\033[m', b, repr(bv)
    else:
        print 'fail', b, a, repr(bv), repr(av)

compare('telepathy.interfaces.CHANNEL', 'TelepathyGLib.IFACE_CHANNEL')
compare('telepathy.interfaces.CHANNEL_INTERFACE_GROUP', 'TelepathyGLib.IFACE_CHANNEL_INTERFACE_GROUP')
compare('telepathy.interfaces.CHANNEL_TYPE_TEXT', 'TelepathyGLib.IFACE_CHANNEL_TYPE_TEXT')
compare('telepathy.interfaces.CHANNEL_TYPE_TUBES', 'TelepathyGLib.IFACE_CHANNEL_TYPE_TUBES')

compare('telepathy.interfaces.CLIENT', 'TelepathyGLib.IFACE_CLIENT')
compare('telepathy.interfaces.CLIENT_HANDLER', 'TelepathyGLib.IFACE_CLIENT_HANDLER')

compare('telepathy.interfaces.CONNECTION', 'TelepathyGLib.IFACE_CONNECTION')
compare('telepathy.interfaces.PROPERTIES_INTERFACE', 'TelepathyGLib.IFACE_DBUS_PROPERTIES')

CONN_INTERFACE_ROOM_CONFIG = \
    'org.freedesktop.Telepathy.Channel.Interface.RoomConfig1'
compare('CONN_INTERFACE_ROOM_CONFIG', 'TelepathyGLib.IFACE_CHANNEL_INTERFACE_ROOM_CONFIG')

compare('telepathy.constants.CHANNEL_GROUP_FLAG_CHANNEL_SPECIFIC_HANDLES', 'TelepathyGLib.ChannelGroupFlags.CHANNEL_SPECIFIC_HANDLES')
compare('telepathy.constants.HANDLE_TYPE_CONTACT', 'TelepathyGLib.HandleType.CONTACT')
compare('telepathy.constants.HANDLE_TYPE_ROOM', 'TelepathyGLib.HandleType.ROOM')
compare('telepathy.constants.PROPERTY_FLAG_WRITE', 'TelepathyGLib.PropertyFlags.WRITE')
```

Aniket21mathur:
@quozl Sorry I disagree with you, you did a small mistake in your code see 
```
exec('av = %s' % a)
    exec('bv = %s' % a)
```
you are comparing 'a' with 'a' that's why all your tests passed. :sweat_smile: .
Although I found out what's the problem is its here 
`compare('telepathy.interfaces.PROPERTIES_INTERFACE', 'TelepathyGLib.IFACE_DBUS_PROPERTIES')`
they are not the same.
`telepathy.interfaces.PROPERTIES_INTERFACE` points to `org.freedesktop.Telepathy.Properties` .
`TelepathyGLib.IFACE_DBUS_PROPERTIES` points to `org.freedesktop.DBus.Properties` .

Almost a week of testing and debugging for that single line of code astonishing and interesting. :smile: .

Aniket21mathur:
After making the above change, following are my observations-

- The alternative that I suggested for DBusProperties as 5cc44e204ebc74ce10c1e1e7b30f1bd1fb547ff1 is working. It might failed due to the incorrect constant or improper testing by me.

- After the last commit collaboration is working. I tested it with the Chat activity. Though I request @pro-panda @chimosky @quozl to test it again. :smile: .

- The next step for us is to find an alternative for `from telepathy.client import Channel`.
Thanks!

quozl:
Thanks for finding my mistake, and for finding the [mistake](https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/401/commits/283dde0de137a1d5f4265bf1b700999c26025ef0#diff-03de8f54944bb6037c70d8ec36f91c15L704) by @pro-panda.  This reminds us always to check everything, even the tools for checking everything.  :grin:

Aniket21mathur:
I worked on replacing `telepathy.client import Channel` , my observations-

- I looked for an alternative class in `TelepathyGlib` found one `TelepathyGlib.Channel.new()` which returned a `TelepathyGlib.Channel` object. But comparing the two objects from telepathy and TelepathyGlib, I found out that the attributes of the two differ a lot, and using `TelepathyGlib.Channel.new()` might involve a large amount of code changes.

- I finally looked into the sources of telepathy and decided to add two classes from the sources `Channel` and `InterfaceFactory` by simplifying their code and removing unwanted variables. I tested the toolkit with the Chat activity, works fine.

- I request further testing, first upto  1207cfe82884d2c03191d24e6b90ad59df24da2d and then upto 
dedb86da4e4c497e4f862a605d8ae12bd0e4bca2 . 

@quozl @chimosky @pro-panda please have a look. Thanks!

Aniket21mathur:
Tested with Maze-activity and calculate activity as well, works fine. @quozl please guide how to proceed 
further. Thanks!

quozl:
Tested dedb86d successfully on two Ubuntu 19.04 virtual machines, with Sugar 0.112 and Chat-85.  Start Chat, share instance, join from neighbourhood view, wait for synchronisation, send text between machines.  No log entries associated with collaboration.  Second virtual machine was created from the first by cloning and then removing `~/.sugar`, restart, then new identities chosen using _My Settings_, _About Me_.

quozl:
Following the above test; when the shared activity was stopped, the neighbourhood view of one of my virtual machines showed _two_ buddy icons for the other machine, and the Chat activity icon.  There should have been only _one_ buddy icon, and the Chat activity icon should have vanished.

The shell.log of the affected machine contains;
```
/usr/lib/python2.7/dist-packages/gi/overrides/Gtk.py:1630: Warning: g_value_transform: assertion 'G_IS_VALUE (src_value)' failed
  return _Gtk_main(*args, **kwargs)
/usr/lib/python2.7/dist-packages/gi/overrides/Gtk.py:1630: Warning: unable to set property 'buddy' of type 'PyObject' from value of type '(null)'
  return _Gtk_main(*args, **kwargs)

(main.py:1876): Gtk-WARNING **: 15:01:41.796: Drawing a gadget with negative dimensions. Did you forget to allocate a size? (node menuitem owner SugarPaletteHeader)
1559714265.591701 ERROR root: set_active() failed: org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying
  File "/usr/local/lib/python2.7/dist-packages/jarabe/desktop/meshbox.py", line 409, in _buddy_removed_cb
    self._remove_buddy(buddy_model)
  File "/usr/local/lib/python2.7/dist-packages/jarabe/desktop/meshbox.py", line 435, in _remove_buddy
    icon = self._buddies[buddy_model.props.key]
KeyError: dbus.ByteArray('AAAAB3NzaC1kc3MAAACBAN0KaucqZ6EoiC8d/Cnnm2jbxCiTDYUpoeNug0nbecNHQdqDRHd9HWGQfKznayOIRbhgDisNQiQA4n0tfVaFUilbo2IJPTPgswAOCXaQXNbRv+YEHi9iAuVjfktu2QAnRAV6KyjbSs/TdXVpFUZcMKDZtdFv18NN6zXHIvaBtuGJAAAAFQC+jn/sPnOxAUITbiOGyyDTw89MeQAAAIEAyYA2fXdouTmdQHL563neQPM8xGGZxxqd+rXU+PQZOfK/1MBpoB+qd6f+ZdNqGudJCWTd3kWi9jWvaVMfAcnUjcpMFESs8dHtWpJ4I3CSI+XdZwZA0C9IrBjVevnFXeigqqWvoERuxUFGEqvdvpZpQnRlC3BE20NCKXS2qADI6+4AAACAa2ia0XqJUs1Gj4HHaZwLbnMAnpO8n70r3lslJp5O3vurLWVRiEOzl+0KJgrABdi7O1yqy0EfoJoAk4MCmcRozCKbCs0OpsrwMaOH0mU5DEIadcETI9AVl3clSyXLBRbV7dwOfKH8I1y+mkJZIkWRAR9DfFkEnBj3eE+TZV88GLw=', variant_level=1)
1559714275.019043 ERROR dbus.connection: Exception in handler for D-Bus signal:
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 230, in maybe_handle_message
    self._handler(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/jarabe/model/neighborhood.py", line 474, in __buddy_activities_changed_cb
    self._update_buddy_activities(buddy_handle, activities)
  File "/usr/local/lib/python2.7/dist-packages/jarabe/model/neighborhood.py", line 536, in _update_buddy_activities
    self._remove_buddy_from_activity(buddy_handle, activity_id)
  File "/usr/local/lib/python2.7/dist-packages/jarabe/model/neighborhood.py", line 553, in _remove_buddy_from_activity
    self._buddy_handles[buddy_handle],
KeyError: dbus.UInt32(2L)
```

The other virtual machine shell.log contains;
```
  File "/usr/local/lib/python2.7/dist-packages/jarabe/desktop/meshbox.py", line 409, in _buddy_removed_cb
    self._remove_buddy(buddy_model)
  File "/usr/local/lib/python2.7/dist-packages/jarabe/desktop/meshbox.py", line 435, in _remove_buddy
    icon = self._buddies[buddy_model.props.key]
KeyError: dbus.ByteArray('AAAAB3NzaC1kc3MAAACBAKKYACo50WaOvhub5ljQYKQas3Qjl26ULFu9dXP028bRHPeUGFg9vG6J4HHlGycmD/TyLai6djPQ4/0SmgKIdfrfTWIDA+QQeAsagsgMQQp8cMBMtPtovUpZvGqMohhsuQ1C+A01kILJ8Pl0InQIbmcxCtNtjj1Demm/pLhTik/9AAAAFQDev4lgG9d9GCTKBcluKxInF/QPFwAAAIB4rCtbNtaKvd4T3Cv4fe0XNvLc9dvzHmuvg3t6Z/eConno3uu/+34tH9ZTI89A8lXINc3nkjX6nzuX7xvWXl8DaakiD7b3vP+Y051qAbdeg3SHBfo7T3pwtQRmGot7nfchQT0uTRCmrJeSufIrbmrYPn4u+DrIn6PZIlq5/vBZJAAAAIB1a59/9o+RbKMvqGceVjh1i6O2aoxCNpmR3eB/yVyZ2WZrsDN4JS/AWptrbLsHmIURV0JPH0iwXM5iuCU6l1yyDG8f7wI6AxV9O6RUMlHtZIPG0dGdYZO5k7nfmzX41xi3AEmBGONTZdvbB210r5c23xVAk53YX6Gpa8QbY85WlA==', variant_level=1)
1559714269.744455 ERROR root: set_active() failed: org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying
```

Aniket21mathur:
Thanks for testing. I tested https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/dedb86da4e4c497e4f862a605d8ae12bd0e4bca2 again using Sugar 0.112 applying patches by hand. I am afraid I didn't got any traceback like yours or faced any problem that you faced, I used Ubuntu 18.04. Made the instances with rdesktop. But after applying patches by hand and saving the changes when I started the chat activity first time It failed, with a traceback 

```
  File "/usr/bin/sugar-activity", line 219, in <module>
    main()
  File "/usr/bin/sugar-activity", line 214, in main
    instance = create_activity_instance(activity_constructor, activity_handle)
  File "/usr/bin/sugar-activity", line 48, in create_activity_instance
    activity = constructor(handle)
  File "/home/guest/Activities/Chat.activity/activity.py", line 86, in __init__
    super(Chat, self).__init__(handle)
  File "/usr/lib/python2.7/dist-packages/sugar3/activity/activity.py", line 428, in __init__
    warn_if_none=False)
  File "/usr/lib/python2.7/dist-packages/sugar3/presence/presenceservice.py", line 88, in get_activity
    dbus_interface=CONN_INTERFACE_ACTIVITY_PROPERTIES)
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 70, in __call__
    return self._proxy_method(*args, **keywords)
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 145, in __call__
    **keywords)
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying
Exited with status 1, pid 2726 data (<open file '<fdopen>', mode 'w' at 0x7f458d704780>, '0e2bd01b1b3900feacbe4107dd5ba8e2c2e61173')

```
I am suprised starting the test activity again after it never caused the traceback, it was  first time starting activity after applying patches that caused it.

quozl:
Note that shell.log is of the yet to be ported Sugar source code, using the static binding.  https://github.com/sugarlabs/sugar/issues/836

quozl:
Your traceback is inconsistent with dedb86da4e4c497e4f862a605d8ae12bd0e4bca2; the line numbers in `presenceservice.py` don't match.

However, the _NoReply_ exception should be understood in the context of the GetActivity call, which is a remote procedure call through D-Bus to Telepathy.  _NoReply_ probably means there was no reply.  Perhaps one of the Telepathy processes died with a segmentation fault or something.  Enable debug logging of Telepathy, and check for interesting crashes in `/var/crash`.  Running the Telepathy processes inside `gdb` is also an effective way to detect and diagnose crashes.  If it doesn't always happen, it will be easier to catch that way.

Aniket21mathur:
> Your traceback is inconsistent with [dedb86d](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/dedb86da4e4c497e4f862a605d8ae12bd0e4bca2); the line numbers in `presenceservice.py` don't match.

Sorry didn't get your point, no changes in presenceservice.py were made, all the changes in https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/dedb86da4e4c497e4f862a605d8ae12bd0e4bca2 were made to presence/activity.py.
 


> Enable debug logging of Telepathy, and check for interesting crashes in `/var/crash`. Running the Telepathy processes inside `gdb` is also an effective way to detect and diagnose crashes. If it doesn't always happen, it will be easier to catch that way.

Thanks. I will do the tests. :-)

chimosky:
> I worked on replacing `telepathy.client import Channel` , my observations-
> 
>     * I looked for an alternative class in `TelepathyGlib` found one `TelepathyGlib.Channel.new()` which returned a `TelepathyGlib.Channel` object. But comparing the two objects from telepathy and TelepathyGlib, I found out that the attributes of the two differ a lot, and using `TelepathyGlib.Channel.new()` might involve a large amount of code changes.


@Aniket21mathur like I said in the email, `TelepathyGlib.Channel` has been replaced with [`TelepathyGLib.SimpleClientFactory.ensure_channel`](https://lazka.github.io/pgi-docs/index.html#TelepathyGLib-0.12/classes/SimpleClientFactory.html#TelepathyGLib.SimpleClientFactory.ensure_channel), same goes for `TelepathyGlib.Connection` which has been replaced with has been replaced with [`TelepathyGlib.SimpleClientFactory.ensure_connection`](https://lazka.github.io/pgi-docs/index.html#TelepathyGLib-0.12/classes/SimpleClientFactory.html#TelepathyGLib.SimpleClientFactory.ensure_connection).
As for the the `immutable_properties` arg, there's no use of `TelepathyGlib.Channel()` in any of the ported files so far, it's used in collabwrapper in the `IncomingFileTransfer` class and one of it's arguments includes a dict of properties which you can use.

quozl:
@Aniket21mathur wrote:
> @quozl wrote:
> > Your traceback is inconsistent with [dedb86d](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/dedb86da4e4c497e4f862a605d8ae12bd0e4bca2); the line numbers in `presenceservice.py` don't match.
> 
> Sorry didn't get your point, no changes in presenceservice.py were made, all the changes in [dedb86d](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/dedb86da4e4c497e4f862a605d8ae12bd0e4bca2) were made to presence/activity.py.

Oh, perhaps you mean that _only_ the changes in the commit dedb86d were applied.  I had assumed that all files as at dedb86d were used.  That's what I did.  So the traceback line numbers are different to the source code because of how you cherry-picked that one set of changes.  I see.  Now that you know what I meant, please reconsider my tests above.

Should you ever need it, below is my script for updating a virtual machine from my desktop git repository for _sugar-toolkit-gtk3_.  SSH must have been already set up using public key methods, and the host discoverable using mDNS (hostname.local) or a local DNS server.  Note that the Python version numbers are assumed, and the target directories are for packaged Sugar, not native.

```sh
#!/bin/sh
HOST=$1
if [ -z $HOST ]; then
    echo no host
    exit 1
fi
shift

PACK=dist-packages
case $HOST in
    fedora*)
        PACK=site-packages
        ;;
esac

rsync -rv --checksum --delete src/sugar3 \
    $HOST:/usr/lib/python2.7/$PACK/
if [ -e bin/sugar-activity3 ]; then
    rsync -rv --checksum --delete src/sugar3 \
        $HOST:/usr/lib/python3/$PACK/
fi

for x in sugar-activity{,3,-web}; do
    if [ -e bin/$x ]; then
        rsync -v bin/$x $HOST:/usr/bin/
        ssh $HOST chmod +x /usr/bin/$x
    fi
done
```

I've another one for _sugar_.

quozl:
@chimosky said
> `TelepathyGlib.Channel` has been replaced with [`TelepathyGLib.SimpleClientFactory.ensure_channel`](https://lazka.github.io/pgi-docs/index.html#TelepathyGLib-0.12/classes/SimpleClientFactory.html#TelepathyGLib.SimpleClientFactory.ensure_channel), same goes for `TelepathyGlib.Connection` which has been replaced with has been replaced with [`TelepathyGlib.SimpleClientFactory.ensure_connection`](https://lazka.github.io/pgi-docs/index.html#TelepathyGLib-0.12/classes/SimpleClientFactory.html#TelepathyGLib.SimpleClientFactory.ensure_connection).
> As for the the `immutable_properties` arg, there's no use of `TelepathyGlib.Channel()` in any of the ported files so far, it's used in collabwrapper in the `IncomingFileTransfer` class and one of it's arguments includes a dict of properties which you can use.

Interesting that a Twisted concept has entered the Telepathy APIs.  I'm looking for example code.  I haven't found any in _sugarlabs_.  Have you ported any activities in this way yet, Ibiam?

(Sugar Labs developers once worked inside Telepathy, GObject and GLib.  Slow withdrawal of our developers from those projects have made them less useful to us.  Perhaps we should be solving this problem by re-engaging?)

Aniket21mathur:
Thanks for the advice @chimosky, I was not very sure about the arguments of the methods you suggested like the immutable_properties, but I will try to use them for sure and share the observations. :-)

Aniket21mathur:
> Oh, perhaps you mean that _only_ the changes in the commit [dedb86d](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/dedb86da4e4c497e4f862a605d8ae12bd0e4bca2) were applied. I had assumed that all files as at [dedb86d](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/dedb86da4e4c497e4f862a605d8ae12bd0e4bca2) were used. That's what I did. So the traceback line numbers are different to the source code because of how you cherry-picked that one set of changes.

Yes, I misunderstood your statement, I thought you only tested the last commit, so applied the patch by hand to sugar 0.112 and tested. I will test it again.

Thanks for the script :-).

chimosky:
> Interesting that a Twisted concept has entered the Telepathy APIs. I'm looking for example code. I haven't found any in sugarlabs. Have you ported any activities in this way yet, Ibiam?

No example code exists yet as no porting has been done, I have a sample while working on telepathy port of collabwrapper. I'll update my repo with some changes and let you know.

chimosky:
@quozl said
> Following the above test; when the shared activity was stopped, the neighbourhood view of one of my virtual machines showed two buddy icons for the other machine, and the Chat activity icon. There should have been only one buddy icon, and the Chat activity icon should have vanished.

Tested dedb86d successfully, after quitting the shared activity it disappears from the neighbourhood leaving just the buddy icon.

Aniket21mathur:
Thanks for testing @chimosky, can you please update us with your testing environment?

quozl:
@chimosky wrote:
> No example code exists yet as no porting has been done, I have a sample while working on telepathy port of collabwrapper. I'll update my repo with the changes and let you know.

Thanks, please do.

We should remember the context; our Port to TelepathyGLib project is replacing one code layer with another; remove the _import telepathy_ (red) in favour of the _TelepathyGLib_ (green);

sugar-collaboration-dependencies

... which results in a huge complexity in integration testing.  Once everything is ported, we'll have to do the real testing.  We need to have ported Sugar, the Toolkit, CollabWrapper, and an activity.  The test activity inside CollabWrapper is my choice.

Aniket21mathur:
@chimosky wrote:


> As for the the `immutable_properties` arg, there's no use of `TelepathyGlib.Channel()` in any of the ported files so far, it's used in collabwrapper in the `IncomingFileTransfer` class and one of it's arguments includes a dict of properties which you can use.

I tried to use these classes as:

```
def __create_text_channel_cb(self, channel_path):
        conn = TelepathyGLib.SimpleClientFactory.ensure_connection(self._connection.object_path)
        self.text_channel = TelepathyGLib.SimpleClientFactory.ensure_channel(conn, channel_path)
        self._tubes_ready()
        
def __create_tubes_channel_cb(self, channel_path):
        conn = TelepathyGLib.SimpleClientFactory.ensure_connection(self._connection.object_path)
        self.tubes_channel = TelepathyGLib.SimpleClientFactory.ensure_channel(conn, channel_path)
        self._tubes_ready()

```
Being not sure I left the immutable_properties argument empty, which resulted in the traceback:

```
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 604, in msg_reply_handler
    reply_handler(*message.get_args_list(**get_args_opts))
  File "/usr/local/lib/python2.7/dist-packages/sugar3/presence/activity.py", line 641, in __create_text_channel_cb
    conn = TelepathyGLib.SimpleClientFactory.ensure_connection(self._connection.object_path)
TypeError: TelepathyGLib.SimpleClientFactory.ensure_connection() takes exactly 3 arguments (1 given)

```
Still looking for an example code of how to use these classes. 

 Looking into https://github.com/sugarlabs/collabwrapper/pull/8/files I found out that the way
`TG.Channel` is used(`channel = TG.Channel(connection.bus_name, object_path)`)  in it will throw out a traceback as we got in https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/401 

 Traceback:
```
  File "/usr/lib/python2.7/site-packages/dbus/connection.py", line 604, in msg_reply_handler
    reply_handler(*message.get_args_list(**get_args_opts))
  File "/usr/lib/python2.7/site-packages/sugar3/presence/activity.py", line 583, in __create_text_channel_cb
    ready_handler=self.__text_channel_ready_cb)
TypeError: GObject.__init__() takes exactly 0 arguments (2 given)
  File "/usr/lib/python2.7/site-packages/dbus/connection.py", line 604, in msg_reply_handler
    reply_handler(*message.get_args_list(**get_args_opts))
  File "/usr/lib/python2.7/site-packages/sugar3/presence/activity.py", line 587, in __create_tubes_channel_cb
    ready_handler=self.__tubes_channel_ready_cb)
TypeError: GObject.__init__() takes exactly 0 arguments (2 given)

```
The traceback is caused as `TG.Channel` do not create a new Channel object, instead it is created by
`TG.Channel.new` .

So I tried it like:

```
def __create_text_channel_cb(self, channel_path):
        conn = TelepathyGLib.Connection.new(TelepathyGLib.DBusDaemon.dup(),self._connection.requested_bus_name, str(self._connection.object_path))
        self.text_channel = TelepathyGLib.Channel.new(conn, channel_path,TelepathyGLib.IFACE_CHANNEL_TYPE_TEXT,TelepathyGLib.HandleType.ROOM,3)
        self._tubes_ready()
        
def __create_tubes_channel_cb(self, channel_path):
        conn = TelepathyGLib.Connection.new(TelepathyGLib.DBusDaemon.dup(),self._connection.requested_bus_name, str(self._connection.object_path))
        self.tubes_channel = TelepathyGLib.Channel.new(conn, channel_path,TelepathyGLib.IFACE_CHANNEL_TYPE_TUBES,TelepathyGLib.HandleType.ROOM,3)
        self._tubes_ready()
```
I successfully got a Channel object as a return, but the methods of that object do not match to those of the object returned by Channel class of telepathy, resulted in tracebacks like 
```
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 604, in msg_reply_handler
    reply_handler(*message.get_args_list(**get_args_opts))
  File "/usr/local/lib/python2.7/dist-packages/sugar3/presence/activity.py", line 650, in __create_tubes_channel_cb
    self._tubes_ready()
  File "/usr/local/lib/python2.7/dist-packages/sugar3/presence/activity.py", line 684, in _tubes_ready
    self._add_self_to_channel()
  File "/usr/local/lib/python2.7/dist-packages/sugar3/presence/activity.py", line 696, in _add_self_to_channel
    group = self.text_channel[TelepathyGLib.IFACE_CHANNEL_INTERFACE_GROUP]
TypeError: 'Channel' object has no attribute '__getitem__'
Exited with status 0, pid 1974 data (<open file '<fdopen>', mode 'w' at 0x7f85d3028780>, 'cb17601232c9818907ba025e5b0f7ba47e30ab96')
```
and 

```
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 230, in maybe_handle_message
    self._handler(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/sugar3/presence/activity.py", line 764, in __text_channel_members_changed_cb
    if TelepathyGLib.IFACE_CHANNEL_INTERFACE_ROOM_CONFIG in self.text_channel:
TypeError: argument of type 'Channel' is not iterable
Exited with status 0, pid 2262 data (<open file '<fdopen>', mode 'w' at 0x7f85d3028780>, 'd72d19cb06f5ca906ad83fc9cd52b8adc4cf92ff')

```
Facing the same issues before forced me two uses the `Channel` and `InterfaceFactory` classes from the telepathy sources itself.
Thanks.

quozl:
I've again looked briefly through the relevant sources with;

```
apt source python-telepathy libtelepathy-glib0 telepathy-salut
```

Please also read https://www.sam.today/blog/telepathy-10-lists.html

By porting to TelepathyGLib, we might be required to change to objects with different methods.  So your observation that returned objects are of a different class is not surprising.

I don't think it is now productive to refer to the CollabWrapper closed pull request, as it never worked.  It may be indicative, and a guide, but as it wasn't finished it is not a verified good source.  Good sources have already been identified as [polari and gnome-shell](https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/401#issuecomment-496322110).

The _immutable properties_ is specified in the documentation to be a GObject.Value (via PyGObject) or a GHashTable (via C function call).  It isn't said to be optional.  I suggest you exercise the interactive REPL to figure out what it wants.  Or use `telepathy-glib/TelepathyGLib-0.12.gir` to read the GIR, where it shows the same types.  Consider the possibility that you've found a mistake in the GIR, as we've seen that kind of thing happen before.

Also consider; step back from the problem and put in words what the yet-to-be-ported code has to do, then go find the way to do it.  When you drill into the detail, it is easy for us to get lost.

Hope that helps.

chimosky:
> Thanks for testing @chimosky, can you please update us with your testing environment?

Testing your branch against a v0.122 instance.

quozl:
> > Thanks for testing @chimosky, can you please update us with your testing environment?
> 
> Testing your branch against a v0.122 instance.

Presume you mean 0.112.  More information needed, please.  Tell us about your testing environment; is it multiple login sessions, or virtual machines?  Are the virtual machines on a bridged network?  What Linux distribution and release?  Is it 0.112 from packages or from source?

Aniket21mathur:
As suggested by @quozl , I looked for crashes that might be causing the No-Reply exception. I did found one in `/var/crash` . I obtained the back trace using `gdb` and `apport-unpack` following the second answer [here](https://askubuntu.com/questions/434431/how-can-i-read-a-crash-file-from-var-crashl).

As,

```
Reading symbols from /usr/lib/telepathy/telepathy-salut...(no debugging symbols found)...done.
[New LWP 593]
[New LWP 597]
[New LWP 758]
[New LWP 760]
[New LWP 829]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/usr/lib/telepathy/telepathy-salut'.
Program terminated with signal SIGTRAP, Trace/breakpoint trap.
#0  0x00007fc5b2f45e11 in ?? () from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
[Current thread is 1 (Thread 0x7fc5b5004940 (LWP 593))]
(gdb) bt
#0  0x00007fc5b2f45e11 in ?? () from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007fc5b2f4715b in g_logv ()
   from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007fc5b2f472af in g_log ()
   from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x00007fc5b2f3eedc in g_source_remove ()
   from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#4  0x000055b5d9dd36a5 in ?? ()
#5  0x00007fc5b321efa3 in g_object_unref ()
   from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#6  0x000055b5d9dd0b17 in ?? ()
#7  0x00007fc5b321efa3 in g_object_unref ()
   from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#8  0x000055b5d9dc9446 in ?? ()
#9  0x00007fc5b321efa3 in g_object_unref ()
   from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#10 0x000055b5d9da5513 in ?? ()
#11 0x00007fc5b321efa3 in g_object_unref ()
   from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#12 0x00007fc5b3242aa0 in g_value_unset ()
   from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#13 0x00007fc5b324363d in g_value_array_free ()
   from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
---Type <return> to continue, or q <return> to quit---
#14 0x00007fc5b3a569c9 in ?? ()
   from /usr/lib/x86_64-linux-gnu/libdbus-glib-1.so.2
#15 0x00007fc5b381fd20 in ?? () from /lib/x86_64-linux-gnu/libdbus-1.so.3
#16 0x00007fc5b3810aea in dbus_connection_dispatch ()
   from /lib/x86_64-linux-gnu/libdbus-1.so.3
#17 0x00007fc5b3a65715 in ?? ()
   from /usr/lib/x86_64-linux-gnu/libdbus-glib-1.so.2
#18 0x00007fc5b2f40387 in g_main_context_dispatch ()
   from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#19 0x00007fc5b2f405c0 in ?? () from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#20 0x00007fc5b2f408d2 in g_main_loop_run ()
   from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#21 0x00007fc5b2c0efe1 in tp_run_connection_manager ()
   from /usr/lib/x86_64-linux-gnu/libtelepathy-glib.so.0
#22 0x000055b5d9d97192 in main ()
```
I am not sure what caused it right now, but is it using telepathy and TelepathyGlib together which is causing the problem? or is there something wrong with TelepathyGLib itself?

Thanks!

quozl:
By classification it is a bug in the Telepathy Salut connection manager; because that is the program that fails.  Can't tell what the immediate cause is without a full trace with function names explained; i.e. one would rerun `gdb` after installing required `-dbg` packages.

But I'm not sure if we should investigate it further.  Suggested preconditions;
* you can easily reproduce the _NoReply_ exception, and;
* you are using either static binding _or_ introspection binding but **not both** in the same process, (using both at once is not the intent of the binding authors), and;
* collaboration isn't working.

Should you use an unported _sugar_ module with a ported _sugar-toolkit-gtk3_ model, you are mixing the bindings.

There's a meme for that.

dcts

Aniket21mathur:
> Should you use an unported _sugar_ module with a ported _sugar-toolkit-gtk3_ model, you are mixing the bindings.

When we test the telepathy ported toolkit, we are running sugar which is having static bindings, so aren't we using the static and introspection bindings together? Is it a good practice to test ported toolkit with sugar having static bindings?

> There's a meme for that.

:laughing: :laughing: .

quozl:
> When we test the telepathy ported toolkit, we are running sugar which is having static bindings, so aren't we using the static and introspection bindings together?

Yes, and we shouldn't; that's why porting Sugar is just as important.

List of toolkit modules that reference Telepathy, with indented list of Sugar shell modules that import them;

* sugar3.activity.activity
  * jarabe.desktop.activitychooser, jarabe.desktop.favoritesview, jarabe.frame.clipboardmenu, jarabe.journal.bundlelauncher, jarabe.journal.bundlelauncher, jarabe.journal.expandedentry, jarabe.journal.iconview, jarabe.journal.journalactivity, jarabe.journal.misc, jarabe.journal.misc, jarabe.model.shell, jarabe.model.shell, jarabe.view.customizebundle, jarabe.view.palettes
* sugar3.presence.presenceservice
  * jarabe.model.filetransfer

You could avoid this by placing the ported toolkit in a subdirectory of the activity, named _sugar3_?

> Is it a good practice to test ported toolkit with sugar having static bindings?

In the same operating system process, no.  But the only way to be sure is to prove it works without any errors.  Putting it another way, when you get an error, or collaboration failure, how can you tell it isn't because of this mixing?

quozl:
Untested speculation; if both the static and the introspection binding ask for replies from Telepathy, when a reply arrives it may be delivered to _one_ of the bindings, and the other binding might report the reply as missing.  I don't know how to test that other than finishing the port.  :grin:

Aniket21mathur:
> Untested speculation; if both the static and the introspection binding ask for replies from Telepathy, when a reply arrives it may be delivered to _one_ of the bindings, and the other binding might report the reply as missing.

Thanks, interesting.

Now I think I might give a reason for the traceback @quozl faced. Although it's just a theory.
The traceback he posted show files only from sugar, having static bindings, no reference of the files from the ported toolkit. It might be a possibility when he stopped the shared activity, the static bindings in sugar didn't got a reply and hence resulted in `org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying`, same might be the reason for the traceback I posted as I stated ~
```
I am suprised starting the test activity again after it never caused the traceback, it was first time starting activity after applying patches that caused it.
```
It is followed by @chimosky testing the toolkit successfully, without facing the problems @quozl faced.
So, it indicates that the ported toolkit have a probability of working which depends upon whether the correct bindings got the reply or not ? 

>  I don't know how to test that other than finishing the port. grin

Neither do I. It's quite interesting as we cannot test anything without porting everything. :stuck_out_tongue:

quozl:
It is common to not be able to test without completing a port.  We have never been able to test GTK 2 and GTK 3 at the same time, or GStreamer 0.10 and GStreamer 1.0.  It is uncommon to be able to mix two APIs.  This situation may have arisen because the Telepathy APIs are shims over D-Bus.  Telepathy mostly executes in separate processes that receive RPC messages from D-Bus.

quozl:
So to summarise the situation on this pull request, both @chimosky and @quozl have tested https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/412/commits/dedb86da4e4c497e4f862a605d8ae12bd0e4bca2 and the behaviour with respect to the shared activity is fine.

Some behaviours were observed that are probably caused by [not yet porting Sugar](https://github.com/sugarlabs/sugar/issues/836), so we can test again one Sugar is ported.

Aniket21mathur:
Thanks. Updated the commit message of the last commit to include reference to the telepathy sources.

chimosky:
> Presume you mean 0.112. More information needed, please? Tell us about your testing environment (for yesterday); was it multiple login sessions, or virtual machines? Were the virtual machines on a bridged network? What Linux distribution and release? Was it 0.112 from packages or from source?

Yeah I meant 0.112, running two sugar build instances in virtual machines and yeah they're both on a bridged network. It was 0.112 from source.

quozl:
@chimosky, thanks.

quozl:
@Aniket21mathur, what would it look like if you used the same mechanic as in https://github.com/sugarlabs/collabwrapper/pull/14/commits/771d87a9482c2783a22c2984f9c9bbc0871edf16?  i.e. replace the imports with constants.

quozl:
In the back of my mind also is how we can avoid the Channel and InterfaceFactory classes, as they exist now because we used the previous API features.  Having looked at the content of the classes, and seeing how little they contain, I'm wondering if it would make sense to remove them and cascade the change to what calls them.  Once we have _sugar_ ported, I hope to have a closer look.

Aniket21mathur:
> @Aniket21mathur, what would it look like if you used the same mechanic as in [sugarlabs/collabwrapper@771d87a](https://github.com/sugarlabs/collabwrapper/commit/771d87a9482c2783a22c2984f9c9bbc0871edf16)? i.e. replace the imports with constants.

Sure, I will do that soon, thanks!

quozl:
On the side-issue of _NoReply_ in shell.log and consequent disappearance of buddy icon, I've reproduced it on 0.114 using static bindings; so I think something we do is triggering a bug in Telepathy Salut.  Let's move that to https://github.com/sugarlabs/sugar/issues/840.

Aniket21mathur:
@quozl I have pushed a commit. In that commit I have removed the use of `dbus` and replaced it with `TelepathyGLib.AccountManager` in file `connectionmanager.py`. Now one thing is insured that we will get a ` TelepathyGLIb.Connection` object instead of `dbus.proxies.ProxyObject` object from connectionmanager.py .

Though it's not yet a totally working system, we still have methods in `presenceservice.py` and `buddy.py` that are not present in Connection class of Telepathy, I am still working on finding an alternative for them. :smile: .

I have commented out the unused code in the commit, so that you can compare.

Thanks!

quozl:
Thanks.  It is very hard to read a commit in unified diff format when you comment out code, so I had to edit a bit before I could figure out what you were thinking of.  I've been reading changes as unified diff for two decades now.  :grin:

The example you refer to is a main program, so it needed a [_GLib.MainLoop_](https://lazka.github.io/pgi-docs/#GLib-2.0/classes/MainLoop.html#GLib.MainLoop) so that the asynchronous method can be used.  In Sugar we have an instance of [_Gtk.main_](https://lazka.github.io/pgi-docs/#Gtk-3.0/functions.html#Gtk.main) which achieves similar outcome.  So I don't think you should be adding a _MainLoop_ to _ConnectionManager_.

But that gives me an idea you may wish to explore; a per-file test.  This one uses [_DBusGMainLoop_](https://dbus.freedesktop.org/doc/dbus-python/dbus.mainloop.html#dbus.mainloop.glib.DBusGMainLoop) as used by `sugar/bin/sugar-install-bundle`, as well as a _GLib.MainLoop_ to manage the timeout.

```diff
--- a/src/sugar3/presence/connectionmanager.py
+++ b/src/sugar3/presence/connectionmanager.py
@@ -118,3 +120,20 @@ def get_connection_manager():
     if not _connection_manager:
         _connection_manager = ConnectionManager()
     return _connection_manager
+
+
+def main():
+    from gi.repository import GLib
+    from dbus.mainloop.glib import DBusGMainLoop
+    DBusGMainLoop(set_as_default=True)
+
+    loop = GLib.MainLoop()
+    cm = get_connection_manager()
+    def test():
+        print cm.get_connections_per_account()
+        return True
+    GLib.timeout_add(1000, test)
+    loop.run()
+
+if __name__ == '__main__':
+    main()
```

Then the `connectionmanager.py` file can be run directly to test it independently.  When run in another desktop, the output is `{}`, but when run inside Sugar Terminal, the output is the Salut Connection object.

Aniket21mathur:
@quozl should I force push the last commit? The changes in the last commit not are required any more as we are not proceeding in that way now?

Aniket21mathur:
@quozl while working on removing the _Channel class, I found out that it will be followed by changes in the activities having collaboration.
I removed the _Channel class using D-Bus directly, tried to test with the Chat activity and got this in logs

```
  File "/home/guest/Activities/Chat.activity/activity.py", line 281, in _shared_cb
    self._setup()
  File "/home/guest/Activities/Chat.activity/activity.py", line 319, in _setup
    self.shared_activity.telepathy_conn)
  File "/home/guest/Activities/Chat.activity/activity.py", line 673, in __init__
    m = self._text_chan[TelepathyGLib.IFACE_CHANNEL].connect_to_signal(
TypeError: 'NoneType' object has no attribute '__getitem__'

```
Indeed we should expect this error because there is no `self._text_chan` object in the toolkit now, being 
replaced by D-Bus.

I am still not understanding why there is a need to remove the _Channel class? _Channel class ultimately uses  dbus object as done in `connectionmanager.py`, self._text_chan is just a way to represent that object. What I think we are doing is changing the architecture without any change in the functionality. Changing this set architecture will demand changes in other activities that involve collaboration.

As per our recent discussion, we can never get away from Dbus, and I don't think we are using anything other than that.

Please guide. :smile: 

Thanks!

quozl:
> @quozl should I force push the last commit? The changes in the last commit not are required any more as we are not proceeding in that way now?

Prefer you `git revert` then `git push`.  Has advantage of letting you write a commit message, and not disrupting anybody who has a copy of your branch with patches on top of it.

quozl:
Code in Chat activity contains _TextChannelWrapper_ which is an earlier or customised copy of the _TextChannelWrapper_ in `collabwrapper.py`.  Maze, Reflect, Speak, Chess and TurtleArt also do this.  Rather than representing a difficulty, it represents work needed to complete the port.  In other words, porting of Chat isn't finished, because the porting of Toolkit isn't finished.

(In the following paragraphs, "package" is a source package maintained by another open source project.  Not a distribution package.)

Restating what we are trying to do; because we wish to port to Python 3, and there is no maintainer for the _telepathy-python_ package on Python 3, we must stop using the _telepathy-python_ package, by using either D-Bus with constants from the _telepathy-glib_ package, or if we really have to, use functions of the _telepathy-glib_ package.

Over ten years, the people who have worked on CollabWrapper, Toolkit, and Chat have mixed together;
* using D-Bus with constants from the _telepathy-python_ package, and
* using functions of the _telepathy-python_ package.

It is this second point above that is the current challenge.  Because the functions and classes of _telepathy-python_ are very different to _telepathy-glib_.

What you have done by adding the _Channel_ class is copy code from the _telepathy-python_ package, rather than port away from the _telepathy-python_ package.  I don't like this because we end up _not porting_, and anybody who is familiar with or tries to use the _telepathy-glib_ package won't be able to understand our code.

Also, the _Channel_ class hasn't really got a good reason to exist.  I learned this while trying to write a class docstring that explains what it is for.  It is an abstraction of a Telepathy channel, without actually being one.

Can we look at it quantitatively as well?  I am expecting changes in other activities; at the least they must port to _telepathy-glib_, and if the toolkit has a **flag day** change to collaboration we will have to version it carefully.

I could more fairly assess the impact if I had two pull requests to compare; one for the scenario where _telepathy-glib_ functions are not used, and one where they are.  You may not have encountered this; that sometimes changes have to be tested and demonstrated to determine if they are to be retained in future.

Aniket21mathur:
> You may not have encountered this; that sometimes changes have to be tested and demonstrated to determine if they are to be retained in future

Thanks for explaining I totally agree with you. :smiley:

I have pushed   64a229689905685822bee6e1514ce39a7fafa1e2, with changes in favour of using dbus.
Some functions were looking significant but were never executed in my tests, so I removed them for now. Also I am not able to understand the significance of `tubes_channel` as it is only declared and never used again in the code, I removed it too. I have also done required changes in chat activity for a one to one connection for testing purpose.

I have tested  64a229689905685822bee6e1514ce39a7fafa1e2 with https://github.com/Aniket21mathur/chat/commit/89c94a3587a443b868d9f09d48bae122a4e032fe using two Oracle machines on bridged network running Ubuntu 18.04.

@quozl please have a look.
Thanks.

Aniket21mathur:
> Prefer you `git revert` then `git push`. Has advantage of letting you write a commit message, and not disrupting anybody who has a copy of your branch with patches on top of it.

Thanks for explaining.

quozl:
Thanks.  Reviewed.  Looks good.

There are some _dbus.Interface_ objects created and destroyed soon after.  Do look out for opportunities to preserve them as class variables, as that will be more efficient.

I did test activity startup.  I did not test collaboration.  I'll wait for corresponding changes to Chat, CollabWrapper, and Sugar.  Please proceed with those.

quozl:
Rebasing this branch onto master is tested.

% git pull --rebase origin master
... resolve a conflict in buddy.py, both imports are to be retained ...
% git rebase --continue

Then

% git push --force.

Let me know if you need further help with that.

I'm rewriting into a private branch at the moment, evaluating each
change in detail.

quozl:
Please review and then cherry-pick (do not merge) commit
https://github.com/quozl/sugar-toolkit-gtk3/commit/191aa9ec38537d8f53f4ffba867c68edbbb13987

I'm now looking through src/sugar3/presence/activity.py

tubes_channel is still used by some activities.  You mentioned a few
days ago that you could not find a reference.  Did you look into
activities?

quozl:
Looking at the branch relative to master, you've removed two other objects
from the public API of presence.Activity; telepathy_text_chan and
telepathy_tubes_chan.  Can these be retained?  That would save on
changes to activities, as there are several references in activities.

quozl:
Please review and cherry-pick
https://github.com/quozl/sugar-toolkit-gtk3/commit/003457f40e51ea5552669af72240de993622eb61

After this, the change vs master is much clearer, concentrating on the
functional changes.

Aniket21mathur:
> ooking at the branch relative to master, you've removed two other objects from the public API of presence.Activity; telepathy_text_chan and telepathy_tubes_chan. Can these be retained?

These are objects returned from telepathy _Channel class and since we are not using that class any more,so I am not sure whether we can retain them or not. How many activities will involve changes? Those which have collaboration or other activities also?

quozl:
On Tue, Jun 25, 2019 at 11:23:34PM -0700, Aniket Mathur wrote:
> These are objects returned from telepathy _Channel class and since
> we are not using that class any more,so I am not sure whether we can
> retain them or not.

It depends on how the activities use the objects.  The Channel class was a
dictionary of D-Bus interfaces.

If we need to, we could use a dictionary of D-Bus interfaces which
isn't the Channel class.

> How many activities will involve changes? Those which have
> collaboration or other activities also?

Not all activities are on GitHub.  Search for the keywords in GitHub
to find some of them.  I've searched my local repositories;

telepathy_tubes_chan; Calculate, CartoonBuilder, CookieSearch,
Dimensions, Distance, FlipSticks, FractionBounce, Jigsaw Puzzle,
Memorize, Play Go, Poll, Portfolio, Read, Readetexts, Record, Reflect,
Show N Tell, Stopwatch, Story, Chess, Turtle Art, Turtle Confusion,
View Slides, and Write activities.

telepathy_text_chan; Browse, Calculate, CartoonBuilder, Chat,
CollabEdit, CookieSearch, Dimensions, Distance, FlipSticks,
FractionBounce, ImageViewer, Jigsaw Puzzle, Level, Maze, Memorize,
Physics, Pippy, Play Go, Poll, Portfolio, Record, Reflect, Show N
Tell, Speak, Stopwatch, Story, Chess, Turtle Art, Turtle Confusion,
and Write activities.

So if we can find a way not to change it, let's do so.

Aniket21mathur:
Thanks. Rebased with master. Reviewed and cherry-picked https://github.com/quozl/sugar-toolkit-gtk3/commit/191aa9ec38537d8f53f4ffba867c68edbbb13987. While reviewing https://github.com/quozl/sugar-toolkit-gtk3/commit/003457f40e51ea5552669af72240de993622eb61 I faced an issue added it as a comment, will cherry-pick and make changes once you approve.

Aniket21mathur:
> Not all activities are on GitHub. Search for the keywords in GitHub to find some of them. I've searched my local repositories;

Looks like a lot of activities use these objects, would be good If we find a way to retain these.

quozl:
Thanks for rebasing.  Much better.

However, there are two commits in our master that are not underneath
this branch.  As if you rebased onto
a7735e6a9714b3ff5a8c36ae92c30fd2dc617339.

Check to see if your local master branch needs updating from
sugarlabs.  If so, "git remote add" and "git pull".

Perhaps my fault for not pointing out that "origin" in my repository
and instructions above was the sugarlabs repository, not my own
repository.  My repository is the remote with name "mine".  Everyone
can use different remote names.

quozl:
Yes, mistake in
https://github.com/quozl/sugar-toolkit-gtk3/commit/003457f40e51ea5552669af72240de993622eb61
fixed in following commit
https://github.com/quozl/sugar-toolkit-gtk3/commit/7de3dd2c59401df855cfd57bc3af1c8bb5091f9a,
thanks for noticing!

Aniket21mathur:
Cherry picked both the commits rebased them into https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/412/commits/31621cfe9e649eaa224cac92f4ceab8d65eb2019. @quozl thanks alot for making these changes. :-)

Aniket21mathur:
> However, there are two commits in our master that are not underneath this branch

Thanks for pointing out, I should have paid attention to that, rebased with sugarlabs/master .

Aniket21mathur:
Did some workaround. An attempt to retain telepathy_tubes_chan and telepathy_text_chan in my case always needs to have something having a structure like the _Channel_ class. But we did the changes favoring dbus to avoid the use of _Channel_ class. Need help in this. 
Thanks!

pro-panda:
> An attempt to retain telepathy_tubes_chan and telepathy_text_chan in my case always needs to have something having a structure like the Channel class

I agree. 
From my memory, I had seen this during my attempt to port to TelepathyGLib and couldn't see any solution. While it'll be a lot of work across activities, I'm fine with moving away from `telepathy_tubes_chan` and `telepathy_text_chan` from activities.

aperezbios:
I found this old documentation, pertaining to how to set up D-Bus tubes, and thought it might be historically useful. See http://wiki.laptop.org/go/Sugar.presence

quozl:
Aniket, don't worry about Channel, as Python is not a strongly typed
language.  What matters is how the objects are used.

I've looked again at activity source code, and here's what activities
rely on;

1.  telepathy_text_chan is to be a dict of at least three D-Bus
interfaces with str keys;

TelepathyGLib.IFACE_CHANNEL aka
'org.freedesktop.Telepathy.Channel'

TelepathyGLib.IFACE_CHANNEL_TYPE_TEXT aka
'org.freedesktop.Telepathy.Channel.Type.Text'

TelepathyGLib.IFACE_CHANNEL_INTERFACE_GROUP aka
'org.freedesktop.Telepathy.Channel.Interface.Group'

2.  telepathy_conn is to be a dict of at least one D-Bus interface
with str key;

TelepathyGLib.IFACE_CONNECTION_INTERFACE_ALIASING aka
'org.freedesktop.Telepathy.Connection.Interface.Aliasing'

3.  telepathy_tubes_chan is to be a dict of at least one D-Bus
interface with str key;

TelepathyGLib.IFACE_CHANNEL_TYPE_TUBES aka
'org.freedesktop.Telepathy.Channel.Type.Tubes'

4.  the D-Bus interfaces may be created at first reference, and
then cached.

There may be other interfaces referenced by some activities, I've not
done a comprehensive search.

Footnote for item 3 above.

At the moment in 0.114 the Toolkit provides a text channel and a tubes
channel.  The text channel is for passing text, or text encoded by the
activity, such as using json.  The tubes channel is for pretty much
anything, even for calling RPC over D-Bus between the activity
instances.

The tubes channel is created using D-Bus interface
org.freedesktop.Telepathy.Channel.Type.Tubes

The interface was deprecated back in June 2012, and a new interface
for tubes is available.

https://telepathy.freedesktop.org/doc/book/sect.tubes.setup.html

However, the old interface is still present in Telepathy, so this
doesn't need to change real soon.  Python 3 is more urgent.

CollabWrapper was to compensate for the deprecation of tubes interface
by switching to the text channel.  Only a few activities were given
the CollabWrapper changes.  Many activities still use tubes.  The
signal NewTube is a good canary.

Some activities that connect to signal NewTube are; Calculate, Cartoon
Builder, Cookie Search, Dimensions, Distance, Flipsticks,
FractionBounce, Jigsaw Puzzle, Memorize, Playgo, Poll, Portfolio,
Read, Readetexts, Record, Reflect, Show N Tell, Stopwatch, Story,
Chess, Turtle Art, Turtle Confusion, View Slides, and Write.

Aniket21mathur:
Thanks for the research, I did some workaround as well:-


> telepathy_text_chan is to be a dict of at least three D-Bus interfaces with str keys;

Yes, agreed. But from the point of the activities only, the toolkit itself uses other keys like _CONN_INTERFACE_ROOM_CONFIG_  and _PROPERTIES_INTERFACE_. There might be activities using other constants as keys.
There might be calls having no keys(there are in the toolkit might be present in the activities as well), as the channel class provides with a default key (_TelepathyGLib.IFACE_CHANNEL_) we might have to handle this case in the activities itself. 

> telepathy_conn is to be a dict of at least one D-Bus interface with str key;

telepathy_conn is a dbus object which the _Activity_ class of `presence/activity.py` takes from `presence/connectionmanager.py`

```
def _track_connection(self, account_path, connection_path):
        connection_name = connection_path.replace('/', '.')[1:]
        bus = dbus.SessionBus()
        connection = bus.get_object(connection_name, connection_path)
        connection.connect_to_signal('StatusChanged',
                                     partial(self.__status_changed_cb,
                                             account_path))
        self._connections_per_account[account_path] = \
            Connection(account_path, connection)

```


> Some activities that connect to signal NewTube are; Calculate, Cartoon Builder, Cookie Search,

Thanks looked at the sources, the uses the NewTube signal, along with self.tubes_chan,  telepathy_tubes_chan is still in use. Sorry I am confused.

quozl:
On Wed, Jun 26, 2019 at 07:54:44PM -0700, Aniket Mathur wrote:
> Thanks for the research, I did some workaround as well:-
> 
>     telepathy_text_chan is to be a dict of at least three D-Bus interfaces with
>     str keys;
> 
> Yes, agreed. But from the point of the activities only, the toolkit itself uses
> other keys like CONN_INTERFACE_ROOM_CONFIG and PROPERTIES_INTERFACE. There
> might be activities using other constants as keys.
> There might be calls having no keys(there are in the toolkit might be present
> in the activities as well), as the channel class provides with a default key (
> TelepathyGLib.IFACE_CHANNEL) we might have to handle this case in the
> activities itself.

Yes, but easy to add those keys.  The default key is more interesting,
but there's only one activity that relies on it as far as I can see.

>     telepathy_conn is to be a dict of at least one D-Bus interface with str
>     key;
> 
> telepathy_conn is a dbus object which the Activity class of presence/
> activity.py takes from presence/connectionmanager.py

Thanks.  Noted.

>     Some activities that connect to signal NewTube are; Calculate, Cartoon
>     Builder, Cookie Search,
> 
> Thanks looked at the sources, the uses the NewTube signal, along with
> self.tubes_chan, telepathy_tubes_chan is still in use. Sorry I am confused.

I've spent a few hours on it just now, see what you think;

https://github.com/quozl/sugar-toolkit-gtk3/commit/1852e981166019eb79dd8c609d502ddb4ddf8807

It almost works, when tested with CollabWrapperTest, just one new
problem appears;

1561605302.537677 ERROR dbus.connection: Exception in handler for D-Bus signal:
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 230, in maybe_handle_message
    self._handler(*args, **kwargs)
  File "/home/guest/Activities/CollabWrapperTest.activity/collabwrapper.py", line 814, in _received_cb
    CHANNEL_TYPE_TEXT].AcknowledgePendingMessages([identity])
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 145, in __call__
    **keywords)
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
DBusException: org.freedesktop.Telepathy.Error.InvalidArgument: invalid message id 0

And I think that's because I'm using CollabWrapper master branch,
without the port to TelepathyGLib.

-- 
James Cameron
http://quozl.netrek.org/

quozl:
> And I think that's because I'm using CollabWrapper master branch,
> without the port to TelepathyGLib.

I've ported CollabWrapper again, and the InvalidArgument does not
change, so I'm looking now at toolkit.  I'm seeing two "finished
setting up tubes" messages, so that might be related.

Aniket21mathur:
> Yes, but easy to add those keys. The default key is more interesting, but there's only one activity that relies on it as far as I can see.

Agreed.

I am still having a look at https://github.com/quozl/sugar-toolkit-gtk3/commit/1852e981166019eb79dd8c609d502ddb4ddf8807. I was trying to do it in a similar manner, but you did it in a more clean way. Thanks.

>  I'm seeing two "finished setting up tubes" messages, so that might be related.

This is because both `__text_get_interfaces_reply_cb` and `__tubes_get_interfaces_reply_cb` calls for self._tubes_read().

quozl:
> This is because both __text_get_interfaces_reply_cb and
> __tubes_get_interfaces_reply_cb calls for self._tubes_read().

Yes.  master branch could also do that, but something prevents it.

To try to find a cause, I've been looking at

git diff master 1852e981166019eb79dd8c609d502ddb4ddf8807 -- src/sugar3/presence/activity.py

but nothing obvious yet.  It is almost as if using the D-Bus API
bypasses something in telepathy-python.

On the invalid message number, my guess is that there's more than one
client as far as D-Bus is concerned; incoming message numbers are
associated with a client, so when the other client asks to acknowledge
one, D-Bus laughs.

quozl:
Found it; self.text_channel and self.tubes_channel are being set much
earlier than each function that calls _tubes_ready.

quozl:
The invalid message number was caused by the double call to
_tubes_ready, once I fixed that double call the error went away.

Fixed in https://github.com/quozl/sugar-toolkit-gtk3/commit/71dac764cac62150dd626cc106724e84e3ea61f2

Tested with CollabWrapper https://github.com/quozl/collabwrapper/commit/0da858c5025a2fbff992c72e01e973350b9e2765

Please review and let me know what you think.

Aniket21mathur:
> Please review and let me know what you think.

Thanks. Reviewed. Looks good.

Should we do the same changes in sugar as well? Using channel and connection dictionaries instead of using a proxy object everytime?

quozl:
> Thanks. Reviewed. Looks good.

Please also test.  My test configuration is a running Sugar from
master branch, then installing the changes to Toolkit and
CollabWrapperTest as above.  The CollabWrapperTest ping and send
buttons are expected to work.

> Should we do the same changes in sugar as well? Using channel and
> connection dictionaries instead of using a proxy object everytime?

Yes.  I'll leave that to you for the moment, so I can work on other
things.  You can use my two patches as examples.  You might also
revert your Chat source code to master; I think that's what I've
tested, and it also works.

quozl:
I've also tested using Telepathy Gabble;

1.  configure each instance with a recognisable public name in My
Settings, About Me,

2.  configure each instance to use jabber.sugarlabs.org in My
Settings, Network, Server,

3.  wait for at least one instance to show a buddy icon for the other,

4.  on the other instance, start CollabWrapperTest, share,

5.  on the first instance, join the shared activity.

Joining shared activity is much faster.

Occasionally the buddy icon for the peer does not appear, but there's
so much reported in shell.log that I think it may be best to wait for
you to work on it.  ;-)

Aniket21mathur:
> Please also test.

Thanks tested with Collabwrapper, Chat and Calculate.

> Yes. I'll leave that to you for the moment, so I can work on other things. You can use my two patches as examples.

Yeah! for sure I will do that. Thanks for helping :-)

Aniket21mathur:
> Occasionally the buddy icon for the peer does not appear, but there's so much reported in shell.log that I think it may be best to wait for you to work on it. ;-)

Yes, I will work on it.

Also tested with https://github.com/sugarlabs/sugar/pull/837/commits/3dcd97a1d295b1f5602f87fa00b71500ebee3dd9.

Aniket21mathur:
> Occasionally the buddy icon for the peer does not appear, but there's so much reported in shell.log that I think it may be best to wait for you to work on it. ;-)

@quozl 
Sorry, tried to reproduce it, never happened to me and hence got nothing in the logs to look into :sweat_smile:

quozl:
Reviewed a7245f9.

There is a change to API, in that `telepathy_text_chan`, a public object, changes type from an instance of _telepathy.InterfaceFactory_ to a dictionary of interfaces.  Functionally similar, but not exactly similar.  The difference is that the caller must now specify `[CHANNEL]` instead of a bare reference.  The object is not documented at the moment, and nor are `telepathy_conn` and `telepathy_tubes_chan`.  The rest of `src/sugar3/presence/activity.py` is also barely documented.  This is an opportunity.  https://github.com/sugarlabs/sugar-toolkit-gtk3/issues/417

PR #415, opened 2019-06-10
Fix check-po target of Makefile and update LINGUAS

Aniket21mathur:


quozl:
Thanks.  Reviewed.  Tested on Ubuntu 18.04.

PR #422, opened 2019-07-25
correct signal name

chimosky:


Aniket21mathur:
Thanks. Reviewed.

chimosky:
@quozl, kindly review

quozl:
Thanks for noticing this.  I've often wondered about it.  I've seen it often.  There are three others to fix in _sugar-toolkit-gtk3_, roughly 15 in _sugar_, and about 200 in the sample of 114 activity repositories on my system.  Can be found with;

```
git grep -E ".connect\('[a-z]*_"
```

The code works.  GLib is happy to match the signal names with underlines.  

In the book Foundations of GTK+ Development, by Andrew Krause;
> When typing the signal name, the underscore and dash characters are interchangeable.  They will be parsed as the same character, so it does not make any difference which one you choose.

Signal names with underscores are aliases.  [g_signal_new](https://developer.gnome.org/gobject/stable/gobject-Signals.html#g-signal-new) does [say](https://github.com/GNOME/glib/blob/master/gobject/gsignal.c#L1401)

> A signal name consists of segments consisting of ASCII letters and digits, separated by either the '-' or '_' character. 

When creating signals for a type, [g_signals_newv](https://github.com/GNOME/glib/blob/master/gobject/gsignal.c#L1720) does automatically duplicate the signal name using underscores in place of dash;

```c
      g_signal_key_bsa = g_bsearch_array_insert (g_signal_key_bsa,
            &g_signal_key_bconfig, &key);
      g_strdelimit (name, "_", '-');
      node->name = g_intern_string (name);
      key.quark = g_quark_from_string (name);
      g_signal_key_bsa = g_bsearch_array_insert (g_signal_key_bsa,
            &g_signal_key_bconfig, &key);
```

When listing available signals for a type, [g_signals_list_ids](https://developer.gnome.org/gobject/stable/gobject-Signals.html#g-signal-list-ids) does [leave out the aliases](https://github.com/GNOME/glib/blob/master/gobject/gsignal.c#L1292).

```c
	/* Signal names with "_" in them are aliases to the same
	 * name with "-" instead of "_".
	 */
	if (!strchr (name, '_'))
	  g_array_append_val (result, keys[i].signal_id);
```

As consequence of the alias automatically created by GLib, and stored in an bsearch array 
data structure indexed by GQuark of the name, which in turn is held in a hashtable, there is no performance advantage to us using dash in place of underscore.

This is a code style issue for us, like flake8.  

Some changes can be too small; it is not worth the `git log` nose for just one.

I'd like to see it done properly.

* [ ] extend the changes to all signal names affected in _sugar-toolkit-gtk3_.

chimosky:
> Signal names with underscores are aliases. g_signal_new does say

Thought as much as it didn't throw any errors.

> This is a code style issue for us, like flake8.
Some changes can be too small; it is not worth the git log nose for just one.
I'd like to see it done properly.

The change was because of our code style, I'll apply the same changes where necessary.

chimosky:
Moved to #423.

PR #423, opened 2019-07-29
Dash in signal names

chimosky:
GLib accepts signal names with dash or underscore and
documents the signals with a dash.

Use of signal names in Sugar was mixed, change to using
dash only.

No functional change.


chimosky:
@quozl kindly review.

PR #432, opened 2019-11-21
Use Gdk.RGBA as Gdk.Color is deprecated

chimosky:
@quozl kindly review.


quozl:
Instantly looks wrong to me, as _Gdk.RGBA_ requires floating point inputs between 0.0 and 1.0 for each colour component, whereas _Gdk.Color_ accepts integers between 0x0 and 0xffff.

PR #439, opened 2020-01-18
Fix icon size in ColorButton [GCI]

srevinsaju:
ColorButton did not show button icon, and swatches were too small
making it almost unusable. Color chooser toolbar had been using
Gtk.IconSize.LARGE_TOOLBAR which return 0 as size.
style.GRID_CELL_SIZE was used as size reference

@quozl @walterbender @Hrishi1999 @Aniket21mathur

### DIFF
--- a/src/sugar3/graphics/colorbutton.py
+++ b/src/sugar3/graphics/colorbutton.py
@@ -66,7 +66,7 @@ class _ColorButton(Gtk.Button):
         self._accept_drag = True
 
         self._preview = Icon(icon_name='color-preview',
-                             pixel_size=style.SMALL_ICON_SIZE)
+                             pixel_size=style.STANDARD_ICON_SIZE)
 
         GObject.GObject.__init__(self, **kwargs)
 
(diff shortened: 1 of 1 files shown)


Hrishi1999:
Tested. Works as expected.

srevinsaju:
Yay. Thanks for the review @Hrishi1999 :)

quozl:
In the video at https://github.com/sugarlabs/terminal-activity/pull/38 the button is larger than a grid cell, causing the height of the new toolbar to be greater than the height of the main toolbar and the edit toolbar.  The button should be the same height and width as all the others.

srevinsaju:
@quozl the video was taken on `sugar-runner`, which respects main screen Gdk.Screen.height and width and not the sub screen. however the grid size remain the same. to see similar change, a test on VM can check if it works

srevinsaju:
basically, what the code does is to use the GRID_CELL_SIZE

srevinsaju:
@quozl , It was my mistake. I have fixed the size issue on  https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/439/commits/daf70d30a44f8697c22a7740c3947c77909a8877. On testing this, on default resolution, I found it was still big. I replaced `GRID_CELL_SIZE` with `STANDARD_ICON_SIZE`, and that fixes the size issue.

quozl:
Thanks.  Tested daf70d3 with Physics and Paint.

Screenshot_udd_2020-01-23_10:09:05

Screenshot_udd_2020-01-23_10:08:50

The light grey circle seems larger than it should be, but I don't know what causes that.  Anyway, it's way better than it was before, so merging.

Fixes https://github.com/sugarlabs/physics/issues/41.

PR #443, opened 2020-02-19
Add --from-code parameter to xgettext for a python3 activity

### LINKED ISSUE #440: For a Python 3 activity, xgettext should be given a better --from-code

chimosky:
Fixes #440, @quozl kindly review.


quozl:
Thanks.  Reviewed.  Is a bit repetitive; any further arguments have to be added in two places.  How about inserting or appending instead?

PR #444, opened 2020-02-27
Build by default for python 3

quozl:
5fa33b2 introduced build for either python 2 or python 3, with the
default of python 2.

`configure` and `autogen.sh` now accept `--with-python3` or
`--with-python2` options, with the default of python 3.

Signed-off-by: James Cameron <quozl@laptop.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -16,16 +16,16 @@ GNOME_COMPILE_WARNINGS(maximum)
 AC_PATH_PROG([GLIB_GENMARSHAL], [glib-genmarshal])
 
 AC_ARG_WITH([python2],
-        AS_HELP_STRING([--with-python2], [build for Python 2 (default)]),
-        [], [with_python2=unset])
+        AS_HELP_STRING([--with-python2], [build for Python 2]),
+        [with_python3=no], [with_python2=unset])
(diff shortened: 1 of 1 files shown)


quozl:
@chimosky, does this build properly on Fedora?

chimosky:
I'll have to wait for this to be merged then patch on fedora and wait for the next compose to test as that's how I'll check for the errors reported.

quozl:
It isn't necessary to wait.  You can change the version in `configure.ac` to end in _~ibiam.0_, then use `make dist` to build a tarball, then use that tarball with `rpmbuild` and a modified spec file.  Sorry, I don't have time to test on Fedora.

Also updated Sugar Live Build.  https://github.com/sugarlabs/sugar-live-build/commit/dbf06fa6631837de09e4fd14ac98928764740f78

chimosky:
I understand what you're saying, I'm saying for the particular error that this is supposed to fix I'll have to wait for a compose as building a spec file won't show me the error.

quozl:
This pull request is not intended to fix any error.

chimosky:
Tested, builds as expected.

PR #445, opened 2020-03-04
Add ctrl y for RedoButton

ryan-van:
Just wanted to add a QOL accelerator for redo button being control + y (this is the usual for many programs).

### DIFF
--- a/src/sugar3/activity/widgets.py
+++ b/src/sugar3/activity/widgets.py
@@ -108,6 +108,7 @@ class RedoButton(ToolButton):
     def __init__(self, **kwargs):
         ToolButton.__init__(self, 'edit-redo', **kwargs)
         self.props.tooltip = _('Redo')
+        self.props.accelerator = '<Ctrl>Y'
 
 
 class CopyButton(ToolButton):
--- a/src/sugar3/activity/widgets.py
+++ b/src/sugar3/activity/widgets.py
@@ -108,6 +108,7 @@ class RedoButton(ToolButton):
     def __init__(self, **kwargs):
         ToolButton.__init__(self, 'edit-redo', **kwargs)
         self.props.tooltip = _('Redo')
+        # From here, redo will be Ctrl + Y to be consistent with Write activity
         self.props.accelerator = '<Ctrl>Y'
 
 


quozl:
Thanks.  Interesting.  By QOL you mean Quality of Life?

I would have thought <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>z</kbd>.  See [Wikipedia Undo](https://en.wikipedia.org/wiki/Undo) and [Ctrl+Y](https://en.wikipedia.org/wiki/Control-Y).

I suggest a user experience design discussion on sugar-devel@ may be helpful.  Let's get more eyes on this.  :eyes: 

I've checked none of the activities I care about use either <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>z</kbd> or <kbd>ctrl</kbd>+<kbd>y</kbd> accelerators.  Pippy tutorial 14 uses <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>z</kbd> for undo, but this is not a functional example.  I was unable to effectively search the other activities, because GitHub search strips punctuation.  I'd have to clone each of them and I don't have the space and funding for that.

Also needs updating on [Wiki Hotkeys](https://wiki.sugarlabs.org/go/Hotkeys) if merged.

ryan-van:
Yea, sorry, I did mean Quality of Life.

It seems like Wikipedia is saying most platforms are now incorporating both `ctrl`+`shift`+`z` and `ctrl`+`y`. And while this would probably be the most ideal, I quickly looked over the ToolButton code, and it may be too complex and not worth the time to build functionality for both of these actions. 

I think from here I can start a discussion on sugar-devel@

Thanks for your input.

quozl:
Thanks for starting the discussion.  It doesn't look like anyone is interested.  If we have no answer after a week, I'll choose a design.

Our Write activity embeds AbiWord which has <kbd>ctrl</kbd>+<kbd>y</kbd> bound to redo.  It is not explicit in the activity, but delegated to the AbiWord widget key bindings.

So my choice in the absence of other opinions is to use <kbd>ctrl</kbd>+<kbd>y</kbd> in order to be consistent with Write activity.

quozl:
Okay, given there are no other opinions, let's go with <kbd>ctrl</kbd>+<kbd>y</kbd>, and mention in the commit messages that we chose that to be consistent with Write activity and we didn't get any other opinions.

ryan-van:
Okay, updated commit and added a comment in the code. Please close the PR if there isn't anything left. Thanks!

quozl:
Thanks.  I've moved the comment into the commit message, as I asked.  I've rebased and squashed, then pushed as 72d36cc964de54403af5257c2b2da81c1e83326e.

quozl:
By the way, I'd normally push changes to your branch, but that doesn't work when your branch is your _master_.  We give guidance already to [use a new branch](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#branching).  Sorry if it wasn't obvious why.  :grin:

PR #449, opened 2020-05-25
Fixes "Logs are not line buffered" #868 issue in sugar repo

free-libre-software:
Reason why it was working in python 2 is because this file is
added only when switched to python 3. Compare
[logger.py](https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/v0.95.2/src/sugar3/logger.py)
file in sugar3 directory of v0.95.2 with
[sugar directory](https://github.com/sugarlabs/sugar-toolkit-gtk3/tree/v0.95.1/src/sugar)
of v0.95.1

Following is not address in this commit from https://github.com/sugarlabs/sugar/issues/868:
> Perhaps what we should be doing now is to configure Sugar and activities for fully buffered logs, and configure line-buffering only when Sugar is being used by a developer. i.e. set `sys.stderr.line_buffering = True` based on an environment variable. Comments?

### DIFF
--- a/src/sugar3/logger.py
+++ b/src/sugar3/logger.py
@@ -171,7 +171,7 @@ def start(log_filename=None):
         try:
             log_path = os.path.join(logs_path, log_filename + '.log')
 
-            log_fd = os.open(log_path, os.O_WRONLY | os.O_CREAT)
+            log_fd = os.open(log_path, os.O_WRONLY | os.O_CREAT, buffer=1)
             os.dup2(log_fd, sys.stdout.fileno())
             os.dup2(log_fd, sys.stderr.fileno())
             os.close(log_fd)


quozl:
Doesn't work.  I get _TypeError: 'buffer' is an invalid keyword argument for this function_.

free-libre-software:
Yes James, you're right, it doesn't work. Sorry for not evaluating thoroughly before PR. In fact original code works fine. I tested it in two scenarios:

1. Running following script both interactively and non-interactively:

```python3
#!/usr/bin/python3

from sugar import logger
import sys
import time

if sys.stdout.isatty():
    string = " Interactive"
else:
    string = " Non-interactive"

logger.start("custom_log")

import logging

while True:
    log_message = time.ctime() + string
    logging.critical(log_message)
    time.sleep(10)
```

2. edit `~/.sugar/default/debug` to uncomment:
````bash
export SUGAR_LOGGER_LEVEL=debug
````

Logs are line buffered in scenario(1) (or not at all buffered). In scenario(2) as well, I see logfile shell.log continuously being populated but that could be due to buffer filling up, given the verbosity.

Can you help me in reproducing #868 and how I can confirm if it's fixed?

quozl:
Sure, it's easy to reproduce.  Find a suitable button activate callback to modify temporarily, add a `logging.error` call to it, restart Sugar, and then prove that _sugar.log_ contain the added message.  Then restart Sugar again, use `tail -f` to watch _sugar.log_ (over SSH or another terminal), then activate the suitable button and verify if you see instant output in _sugar.log_, or if the output is delayed until Sugar is stopped.

https://github.com/sugarlabs/sugar/issues/868 has an explanation and a solution "`sys.stderr.line_buffering = True`" that seems like it should work.  Have you tried it?

free-libre-software:
> Sure, it's easy to reproduce. Find a suitable button activate callback to modify temporarily, add a `logging.error` call to it, restart Sugar, and then prove that _sugar.log_ contains the added message. Then restart Sugar again, use `tail -f` to watch _sugar.log_ (over SSH or another terminal), then activate the suitable button and verify if you see instant output in _sugar.log_, or if the output is delayed until Sugar is stopped.
> 
> [sugarlabs/sugar#868](https://github.com/sugarlabs/sugar/issues/868) has an explanation and a solution "`sys.stderr.line_buffering = True`" that seems like it should work. Have you tried it?

Thanks James, I am seeing logs instantly. Perhaps you can give me specific steps to follow to reproduce this error.

I made following changes to `/usr/src/sugar/src/jarabe/model/sound.py`:
* `import logging`
* In PlaybackSound.set_muted() inserted `logging.error('xyzabc')`

You can see the output below:

demo

quozl:
Thanks, I agree.

Closing as _Not Sugar_.

Problem is solved by upgrading to Python 3.8.2, though there may be an earlier version that had the fix.

PR #454, opened 2020-12-02
Add setup.py --destdir support

quozl:
* simplify messages,
* add desktop and appdata files to messages,
* add DESTDIR support.

### DIFF
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -415,6 +415,7 @@ class Installer(Packager):
             os.makedirs(os.path.dirname(path))
         tree = ET.ElementTree(root)
         tree.write(path, encoding='UTF-8')
+        print('Install %s' % (path))
 
 
 def cmd_check(config, options):
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -280,7 +280,8 @@ class Installer(Packager):
         Packager.__init__(self, builder.config)
         self.builder = builder
 
-    def install(self, prefix, install_mime=True, install_desktop_file=True):
+    def install(self, destdir, prefix,
+                install_mime=True, install_desktop_file=True):
         self.builder.build()
 
         activity_path = os.path.join(prefix, 'share', 'sugar', 'activities',
@@ -290,16 +291,21 @@ class Installer(Packager):
 
         for f in self.get_files_in_git():
             source_path = os.path.join(self.config.source_dir, f)
-            dest_path = os.path.join(activity_path, f)
+            dest_path = os.path.join(destdir,
+                                     os.path.relpath(activity_path, '/'), f)
             source_to_dest[source_path] = dest_path
 
         for f in self.builder.get_locale_files():
             source_path = os.path.join(self.builder.locale_dir, f)
 
             if source_path.endswith(".mo"):
-                dest_path = os.path.join(prefix, 'share', 'locale', f)
+                dest_path = os.path.join(destdir,
+                                         os.path.relpath(prefix, '/'),
+                                         'share', 'locale', f)
             else:
-                dest_path = os.path.join(activity_path, 'locale', f)
+                dest_path = os.path.join(destdir,
+                                         os.path.relpath(activity_path, '/'),
+                                         'locale', f)
 
             source_to_dest[source_path] = dest_path
 
@@ -316,10 +322,10 @@ class Installer(Packager):
             self.config.bundle.install_mime_type(self.config.source_dir)
 
         if install_desktop_file:
-            self._install_desktop_file(prefix, activity_path)
-            self._generate_appdata(prefix, activity_path)
+            self._install_desktop_file(destdir, prefix, activity_path)
+            self._generate_appdata(destdir, prefix, activity_path)
 
-    def _install_desktop_file(self, prefix, activity_path):
+    def _install_desktop_file(self, destdir, prefix, activity_path):
         cp = ConfigParser()
         section = 'Desktop Entry'
         cp.add_section(section)
@@ -327,7 +333,9 @@ class Installer(Packager):
 
         # Get it from the activity.info for the non-translated version
         info = ConfigParser()
-        info.read(os.path.join(activity_path, 'activity', 'activity.info'))
+        info_path = os.path.join(destdir, os.path.relpath(activity_path, '/'),
+                                 'activity', 'activity.info')
+        info.read(info_path)
         cp.set(section, 'Name', info.get('Activity', 'name'))
         if info.has_option('Activity', 'summary'):
             cp.set(section, 'Comment', info.get('Activity', 'summary'))
@@ -353,16 +361,19 @@ class Installer(Packager):
         cp.set(section, 'Path', activity_path)  # Path == CWD for running
 
         name = '{}.activity.desktop'.format(self.config.bundle_id)
-        path = os.path.join(prefix, 'share', 'applications', name)
+        path = os.path.join(destdir, os.path.relpath(prefix, '/'),
+                            'share', 'applications', name)
         if not os.path.isdir(os.path.dirname(path)):
             os.makedirs(os.path.dirname(path))
         with open(path, 'w') as f:
             cp.write(f)
         print('Install %s' % (path))
 
-    def _generate_appdata(self, prefix, activity_path):
+    def _generate_appdata(self, destdir, prefix, activity_path):
         info = ConfigParser()
-        info.read(os.path.join(activity_path, 'activity', 'activity.info'))
+        info_path = os.path.join(destdir, os.path.relpath(activity_path, '/'),
+                                 'activity', 'activity.info')
+        info.read(info_path)
 
         required_fields = ['metadata_license', 'license', 'name', 'icon',
                            'description']
@@ -409,7 +420,8 @@ class Installer(Packager):
             ET.SubElement(root, 'url', type='bugtracker').text = \
                 info.get('Activity', 'repository')
 
-        path = os.path.join(prefix, 'share', 'metainfo',
+        path = os.path.join(destdir, os.path.relpath(prefix, '/'),
+                            'share', 'metainfo',
                             self.config.bundle_id + '.appdata.xml')
         if not os.path.isdir(os.path.dirname(path)):
             os.makedirs(os.path.dirname(path))
@@ -507,6 +519,7 @@ def cmd_install(config, options):
 
     installer = Installer(Builder(config))
     installer.install(
+        options.destdir,
         options.prefix,
         options.install_mime,
         options.install_desktop_file)
@@ -594,6 +607,9 @@ def start():
     install_parser.add_argument(
         "--prefix", dest="prefix", default=sys.prefix,
         help="Path for installing")
+    install_parser.add_argument(
+        "--destdir", dest="destdir", default="/",
+        help="Path for staged install")
     install_parser.add_argument(
         "--skip-install-mime", dest="install_mime",
         action="store_false", default=True,
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -304,7 +304,7 @@ class Installer(Packager):
             source_to_dest[source_path] = dest_path
 
         for source, dest in list(source_to_dest.items()):
-            print('Install %s to %s.' % (source, dest))
+            print('Install %s' % (dest))
 
             path = os.path.dirname(dest)
             if not os.path.exists(path):
--- a/src/sugar3/activity/bundlebuilder.py
+++ b/src/sugar3/activity/bundlebuilder.py
@@ -358,6 +358,7 @@ class Installer(Packager):
             os.makedirs(os.path.dirname(path))
         with open(path, 'w') as f:
             cp.write(f)
+        print('Install %s' % (path))
 
     def _generate_appdata(self, prefix, activity_path):
         info = ConfigParser()


chimosky:
Tried testing but I don't notice any changes, testing currently without supplying `--prefix` to master installs to `/usr/share/sugar/activities/activityname.activity/` and same also happens for this branch.

How did you test?

quozl:
Thanks for testing.  That's expected.  My test method was;

* prepare a system with Ubuntu 20.04 and Sugar 0.117, and apply the patch,
* manually remove the files that had been installed by the `sugar-calculate-activity` package, but without removing the package,
```log
rm -rvf \
/usr/share/applications/org.laptop.Calculate.activity.desktop \
/usr/share/doc/sugar-calculate-activity \
/usr/share/locale/*/LC_MESSAGES/org.laptop.Calculate.mo \
/usr/share/sugar/activities/Calculate.activity
```
* verify the activity is absent from the Home View,
* clone the Calculate activity and patch metadata so that the appdata code path is tested;
```diff
--- a/activity/activity.info
+++ b/activity/activity.info
@@ -10,3 +10,5 @@ summary = This ...
 repository = https://github.com/sugarlabs/sugarlabs-calculate
 url = https://help.sugarlabs.org/calculate.html
 tags = Maths
+metadata_license = foo
+description = bar
```
* test _install_ step without options, with `--prefix`, with `--destdir`, and with both, each time verifying the output.

e.g.
```log
python3 setup.py install
python3 setup.py install --prefix /usr/local
python3 setup.py install --prefix /usr --target /tmp/prefix
```

Expected output;

| `--prefix` | `--destdir` | destinations |
| ---- | ---- | :---- |
| omitted | omitted |  `/usr/share/locale/*/LC_MESSAGES/org.laptop.Calculate.mo`<br>`/usr/share/sugar/activities/Calculate.activity`<br>`/usr/share/applications/org.laptop.Calculate.activity.desktop`<br>`/usr/share/metainfo/org.laptop.Calculate.appdata.xml` |
| `/usr/local` | omitted | `/usr/local/share/sugar/activities/Calculate.activity/locale/hy/activity.linfo`<br>`/usr/local/share/locale/hy/LC_MESSAGES/org.laptop.Calculate.mo`<br>`/usr/local/share/applications/org.laptop.Calculate.activity.desktop`<br>`/usr/local/share/metainfo/org.laptop.Calculate.appdata.xml` |
| `/usr` | `/tmp/prefix` | `/tmp/prefix/usr/share/locale/*/LC_MESSAGES/org.laptop.Calculate.mo`<br>`/tmp/prefix/usr/share/sugar/activities/Calculate.activity`<br>`/tmp/prefix/usr/share/applications/org.laptop.Calculate.activity.desktop`<br>`/tmp/prefix/usr/share/metainfo/org.laptop.Calculate.appdata.xml` |

* removing all files added by the test, and then reinstalling the `sugar-calculate-activity` package,
```log
rm -rvf /usr/share/applications/org.laptop.Calculate.activity.desktop \
/usr/share/doc/sugar-calculate-activity \
/usr/share/locale/*/LC_MESSAGES/org.laptop.Calculate.mo \
/usr/share/sugar/activities/Calculate.activity \
/usr/share/metainfo/org.laptop.Calculate.appdata.xml
rm -rvf /usr/local/share/applications/org.laptop.Calculate.activity.desktop \
/usr/local/share/doc/sugar-calculate-activity \
/usr/local/share/locale/*/LC_MESSAGES/org.laptop.Calculate.mo \
/usr/local/share/sugar/activities/Calculate.activity \
/usr/local/share/metainfo/org.laptop.Calculate.appdata.xml
rm -rvf /tmp/prefix
apt install --reinstall sugar-calculate-activity
```
* verify the activity is present in the Home View,
* reverse the patch.

chimosky:
Testing on live build and when I test with `--prefix` set to `/usr/local` the activity files get installed to `/usr/local` and trying to install with `--destdir` throws an error `./setup.py: error: unrecognized arguments: --destdir /tmp/target` and when the activity is installed with `--prefix` set to `/usr/local`, the desktop file doesn't exist in `/usr/share/applications/`.

quozl:
An unrecognised argument for `--destdir` means my patch isn't applied to the bundle builder being used.  There's more than one;
* the Sugar Toolkit for GTK 2,
* the Sugar Toolkit for GTK 3 on Python 2,
* the Sugar Toolkit for GTK 3 on Python 3.

The `exec` shows the source you are using is for the Sugar Toolkit for GTK 3 on Python 2, and you should be able to confirm that by reading the first line of setup.py.  Where did the source code come from?  If it is a git repository, what's the git hash?

quozl:
Downside of 8250c17 and f9b2b6c is withdrawal of Python 2 support by the toolkit.  Comments?

walterbender:
I think it is time. It is enough work to support the Python 3 code given our limited resources.

quozl:
It will stop further porting to Python 3.  That's also a good idea, because it will ensure our smaller number of activities will better match our limited resources.

srevinsaju:
So, does this mean, we stop migrating to python3?

quozl:
By withdrawing the toolkit for Python 2, we reduce availability of the environment used for porting, such as Sugar Live Build or Ubuntu.  Porting could still occur, but is made more difficult because the environment used for porting would be ageing rapidly.

Putting it differently, are your changes in 8250c17 and f9b2b6c sufficient enough to justify this impact.  I suspect not.  So I'm looking for further input.

srevinsaju:
> By withdrawing the toolkit for Python 2, we reduce availability of the environment used for porting, such as Sugar Live Build or Ubuntu. Porting could still occur, but is made more difficult because the environment used for porting would be ageing rapidly.

> Putting it differently, are your changes in [8250c17](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/8250c1785d881f5e711c9a1362034d539ef390cd) and [f9b2b6c](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/f9b2b6c9f465dd816121b15ca625a5c775661f7e) sufficient enough to justify this impact. I suspect not. So I'm looking for further input.


It is quite complex for me to understand. Why do we stop migrating to Python 3? Are we EOL'ing Sugar? This would mean lesser outreach, distribution I suppose. I did not get the objective :(

quozl:
Happy to explain again.

Porting an activity from [Python 2 to Python 3](https://github.com/sugarlabs/sugar-docs/blob/master/src/python-porting-guide.md) requires both environments to be working, otherwise the developer can't tell if program behaviours they observe are correctly implemented.  We had one or two people try to port without having a Python 2 environment available, and it has been a terrible outcome, because they introduce bugs, regressions, and unexpected features.  They end up relying on other people to notice these, because without a Python 2 environment they can't see the difference.

My objective had been to support both Python 2 and Python 3 in the Sugar Toolkit, see https://github.com/sugarlabs/sugar-toolkit-gtk3/issues/382.  Your objective in 8250c17 is to increase performance by removing a check that is now available in Python 3, but the code you've changed fails in Python 2.

So we have conflicting objectives, and I'm asking for developers to decide which objective takes precedence.

srevinsaju:
> Happy to explain again.
> 
> Porting an activity from [Python 2 to Python 3](https://github.com/sugarlabs/sugar-docs/blob/master/src/python-porting-guide.md) requires both environments to be working, otherwise the developer can't tell if program behaviours they observe are correctly implemented. We had one or two people try to port without having a Python 2 environment available, and it has been a terrible outcome, because they introduce bugs, regressions, and unexpected features. They end up relying on other people to notice these, because without a Python 2 environment they can't see the difference.
> 
> My objective had been to support both Python 2 and Python 3 in the Sugar Toolkit, see #382. Your objective in [8250c17](https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/8250c1785d881f5e711c9a1362034d539ef390cd) is to increase performance by removing a check that is now available in Python 3, but the code you've changed fails in Python 2.
> 
> So we have conflicting objectives, and I'm asking for developers to decide which objective takes precedence.

Oh I see, I understood it now. Thanks a lot for explaining it again for me :smile: :sweat_smile:

quozl:
Thanks.  I've removed the commits.  What information do you have to suggest Python 3 is higher performance for Sugar and activities?  In my own tests, Python 3 shows lower performance, which I presume is due to the much larger functionality of Python 3.

chimosky:
@quozl said 
> An unrecognised argument for --destdir means my patch isn't applied to the bundle builder being used. There's more than one;

Yes that's how it should be but this happens, your patch was applied and I confirmed it by looking at `/usr/lib/python3.7/site-packages/sugar3/activity/bundlebuilder.py` and the changes are there.

>The exec shows the source you are using is for the Sugar Toolkit for GTK 3 on Python 2, and you should be able to confirm that by reading the first line of setup.py. Where did the source code come from? If it is a git repository, what's the git hash?

A reinstall shows `Exec` as Sugar Toolkit for GTK3 on Python 2, an install with `python3 setup.py install` shows `Exec` as for Sugar Toolkit for GTK3 on Python3 which is expected but I don't understand why a reinstall automatically installs for Python 2.
Although it seems to be installing v44 which is shown in the build path that's embedded in the .desktop file after reinstalling where the latest version is v46 which is shown by the debian [changelog](https://salsa.debian.org/pkg-sugar-team/sugar-calculate-activity/-/blob/master/debian/changelog).
`Icon = /build/sugar-calculate-activity-44/debian/sugar-calculate-activity//usr/share/sugar/activities/Calculate.activity/activity/calculate.svg`

chimosky:
> Downside of 8250c17 and f9b2b6c is withdrawal of Python 2 support by the toolkit. Comments?

Maintaining two separate branches might be good but it's extra work, although we can keep Python 2 images on sunjammer for anyone who needs them. I don't think withdrawing support for Python 2 is great as we still have activities that haven't been ported to Python 3 yet.

quozl:
> @quozl said
> 
> > An unrecognised argument for --destdir means my patch isn't applied to the bundle builder being used. There's more than one;
> 
> Yes that's how it should be but this happens, your patch was applied and I confirmed it by looking at `/usr/lib/python3.7/site-packages/sugar3/activity/bundlebuilder.py` and the changes are there.

You may apply the patch to the other instances of bundlebuilder.py if you wish.  That should entirely fix the unrecognised argument symptom.

> > The exec shows the source you are using is for the Sugar Toolkit for GTK 3 on Python 2, and you should be able to confirm that by reading the first line of setup.py. Where did the source code come from? If it is a git repository, what's the git hash?
> 
> A reinstall shows `Exec` as Sugar Toolkit for GTK3 on Python 2, an install with `python3 setup.py install` shows `Exec` as for Sugar Toolkit for GTK3 on Python3 which is expected but I don't understand why a reinstall automatically installs for Python 2.
> Although it seems to be installing v44 which is shown in the build path that's embedded in the .desktop file after reinstalling where the latest version is v46 which is shown by the debian [changelog](https://salsa.debian.org/pkg-sugar-team/sugar-calculate-activity/-/blob/master/debian/changelog).
> `Icon = /build/sugar-calculate-activity-44/debian/sugar-calculate-activity//usr/share/sugar/activities/Calculate.activity/activity/calculate.svg`

You're trying to test the bundle builder against an activity source.  Where did the source come from? If it is a git repository, what's the git hash?  While activity versions can be indicative, you've mentioned two, and you've mentioned derivative source (by Debian project), so I'm not sure what you are using.  I also don't know what you are talking about with reinstall; I only mentioned reinstall in my instructions above in order to restore a system to the state it was in before using it for testing.  The bundle builder is not involved in Debian package installation.

quozl:
> Maintaining two separate branches might be good but it's extra work, although we can keep Python 2 images on sunjammer for anyone who needs them. I don't think withdrawing support for Python 2 is great as we still have activities that haven't been ported to Python 3 yet.

Thanks.  Don't worry about the extra work; I already maintain the Fedora 18 branch for all of Sugar and activities under OLPC OS 13.x, and packaging branches for OLPC OS 18 and 20, so if Sugar Labs decides to drop Python 2 support what it really means is that I'll maintain it alone, as I already do for the other branches.  Dropping Python 2 support would be a decision by Sugar Labs _not to maintain_, leaving it to others to do so.

quozl:
@srevinsaju, did you test this?  Did it work for you?  Any code review?

chimosky:
> You may apply the patch to the other instances of bundlebuilder.py if you wish. That should entirely fix the unrecognised argument symptom.

I didn't need to at the time as it had already been applied when I built from this branch.

> You're trying to test the bundle builder against an activity source. Where did the source come from? If it is a git repository, what's the git hash? 

At the time it was from a git repository, with HEAD at sugarlabs/calculate-activity@d36f122.


> While activity versions can be indicative, you've mentioned two, and you've mentioned derivative source (by Debian project), so I'm not sure what you are using. I also don't know what you are talking about with reinstall; I only mentioned reinstall in my instructions above in order to restore a system to the state it was in before using it for testing. The bundle builder is not involved in Debian package installation, so any behaviours you observed with a reinstall are tangential.

The source of the activity was a git repository, I shouldn't have mentioned the derivative source but I'd checked and the versions were different which wasn't a surprise as I hadn't updated my test environment at the time; yes I don't expect you to know about the reinstall issue as that's also debian related.

I'll test again and let you know when I do.

quozl:
Thanks, please do.  The data so far conflicts; https://github.com/sugarlabs/calculate-activity/commit/d36f122 has a `setup.py` of

```python
#!/usr/bin/env python
from sugar3.activity import bundlebuilder
bundlebuilder.start()
```

yet you got `./setup.py: error: unrecognized arguments: --destdir`.  Perhaps you were testing on a system where `/usr/bin/python` is Python 2, and you used `./setup.py` instead of `python3 setup.py`, which implies Python 2.  I note that the Debian packaging for the Calculate activity;

* https://salsa.debian.org/pkg-sugar-team/sugar-calculate-activity/-/blob/master/debian/rules#L3 uses cdbs, and;
* https://salsa.debian.org/build-common-team/cdbs/-/blob/master/1/class/python-sugar.mk.in#L119 explicitly calls `setup.py` from within `python3` not `python`; it ignores what `setup.py` specifies.

We can no longer assume `/usr/bin/python` to be a specific version of Python, now that downstream distributions have diverged on that.  In my cache of about 163 activity repositories, there's a variation in how the Python interpreter is chosen;

| `#!` in setup.py | frequency |
| -- | -- |
| `/usr/bin/python2` | 1 |
| `/usr/bin/python3` | 4 |
| `/usr/bin/python` | 16 |
| `/usr/bin/env python2` | 0 |
| `/usr/bin/env python3` | 28 |
| `/usr/bin/env python` | 108 |

chimosky:
> yet you got `./setup.py: error: unrecognized arguments: --destdir`. Perhaps you were testing on a system where `/usr/bin/python` is Python 2, and you used `./setup.py` instead of `python3 setup.py`, which implies Python 2. I note that the Debian packaging for the Calculate activity;

All tests were done with `python3 setup.py` and I just did it again and I still get same error in same format, I can't debug more at the moment to find out why.

> We can no longer assume /usr/bin/python to be a specific version of Python, now that downstream distributions have diverged on that.

Agreed.

quozl:
The patch adds `--destdir` support.  Therefore the patch must not be applied.  Look for evidence of the feature in the bundlebuilder.py file that is executed.  You might use strace to prove which file is executed.  If you have compiled Python files, make sure to recompile them.

srevinsaju:
> @srevinsaju, did you test this? Did it work for you? Any code review?

Nope, haven't tested this yet. Will surely test it!

quozl:
Rebased branch after 0.118 release.

srevinsaju:
Tested, works as expected.

srevinsaju (approved):

chimosky:
Tested, works as expected. 

I couldn't test earlier as I forgot to copy the contents of `/usr/lib/python3.7/site-packages/sugar3` to `/usr/lib/python3.7/dist-packages`.

PR #455, opened 2021-01-06
ci: automatically deploy documentation to gh-pages

srevinsaju:
```
ci: install sphinx
ci: deploy to sugar3 subdirectory
ci: add deploy/.nojekyll
ci: first make sugar-toolkit-gtk3 before docs
ci: enable deb-src repositories
ci: add gir1.2-sugarext-1.0 and remove sugar build
```

## Procedure
* Merge this pull request after reviewing the difference between https://developer.sugarlabs.org/sugar3 and https://dev.sl.srevinsaju.me/sugar3
* Ping me on IRC :smile:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,45 @@
+
+name: Documentation
+on:
+  push:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        
+      - name: Prepare apt repositories
+        run: | 
+          sudo sed -i '/deb-src/s/^# //' /etc/apt/sources.list
+          sudo apt update
+
+      - name: Install sugar-toolkit-gtk3
+        run: |
+          sudo apt build-dep sugar-toolkit-gtk3
+          sudo apt install python-is-python3 python3-cairo python3-gi python3-gi-cairo python3-dateutil python3-decorator \
+            gir1.2-telepathyglib-0.12 gir1.2-webkit2-4.0 python3-empy gir1.2-sugarext-1.0
+          ./autogen.sh --with-python2
+          make
+          ./autogen.sh --with-python3
+          make
+          
+      - name: Install Sphinx
+        run: |
+          sudo apt-get -qqy install sphinx-common
+          
+      - name: Build docs
+        run: |
+          ./make-doc.sh
+          mkdir deploy
+          mv doc/_build/html deploy/sugar3
+          touch deploy/.nojekyll
+
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@3.7.1
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH: gh-pages # The branch the action should deploy to.
+          FOLDER: deploy  # The folder the action should deploy.
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,40 +2,35 @@
 name: Documentation
 on:
   push:
-    branches: [ master ]
   workflow_dispatch:
 
 jobs:
   build:
     runs-on: ubuntu-20.04
-    steps:
-      - uses: actions/checkout@v2
-        
-      - name: Prepare apt repositories
-        run: | 
-          sudo sed -i '/deb-src/s/^# //' /etc/apt/sources.list
-          sudo apt update
+    container:
+      image: 'archlinux:latest'
 
-      - name: Install sugar-toolkit-gtk3
+    steps:
+      - name: Install dependencies
         run: |
-          sudo apt build-dep sugar-toolkit-gtk3
-          sudo apt install python-is-python3 python3-cairo python3-gi python3-gi-cairo python3-dateutil python3-decorator \
-            gir1.2-telepathyglib-0.12 gir1.2-webkit2-4.0 python3-empy gir1.2-sugarext-1.0
-          ./autogen.sh --with-python2
-          make
-          ./autogen.sh --with-python3
-          make
-          
-      - name: Install Sphinx
+          # assume yes and do not ask for confirmation when installing deps using pacman
+          yes | pacman -Syu base-devel sudo git wget curl python-sphinx rsync --noconfirm
+
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+
+      - name: Create builduser
         run: |
-          sudo apt-get -qqy install sphinx-common
-          
-      - name: Build docs
+          # A user is required so that the docs are not built as root
+          useradd builduser -m # Create the builduser
+          passwd -d builduser # Delete the buildusers password
+          printf 'builduser ALL=(ALL) ALL\n' | tee -a /etc/sudoers
+
+      - name: Build documentation
         run: |
-          ./make-doc.sh
-          mkdir deploy
-          mv doc/_build/html deploy/sugar3
-          touch deploy/.nojekyll
+          sudo chown builduser:builduser . -R
+          sudo -u builduser ./ci/docs.sh
 
       - name: Deploy 🚀
         uses: JamesIves/github-pages-deploy-action@3.7.1
--- a/ci/docs.sh
+++ b/ci/docs.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+# Helper script to build documentation using Sphinx
+# NOTE: Should only be run on archlinux:latest 
+
+GREEN='\033[0;32m'
+NC='\033[0m'
+
+set -eux
+
+show-green () {
+    echo -e "${GREEN} ==> ${1} ${NC}" 
+}
+
+# build package from the AUR (arch user repository)
+if command -v makepkg &> /dev/null
+then
+    show-green "Cloning sugar-toolkit-gtk3-git from AUR"
+    mkdir .cache
+    cd .cache 
+    git clone https://aur.archlinux.org/sugar-toolkit-gtk3-git.git
+    cd sugar-toolkit-gtk3-git
+    makepkg -sif --noconfirm 
+    show-green "Installing optional dependencies"
+    yes | sudo pacman -S --needed $(cat .SRCINFO| grep 'optdepends' | grep -o '= .*:' | sed 's,= ,,g' | sed 's,:,,g') --noconfirm
+    cd ../..
+    show-green "Dependencies synced, packages built"
+
+fi
+
+
+# make source
+show-green "Compiling"
+./autogen.sh --with-python3
+make
+
+# make documentation
+show-green "Building documentation"
+./make-doc.sh
+mkdir deploy
+mv doc/_build/html deploy/sugar3
+touch deploy/.nojekyll
+# create an index.html so that users dont become confused
+show-green "Writing index.html"
+echo "<h1>Page Moved</h1>" > index.html
+echo "<p>We have moved this page to <a href=\"https://github.com/sugarlabs/sugar-docs/blob/master/README.md\">GitHub</a>.</p>" >> index.html
+echo "<p>How did you get here? Please <a href=\"https://github.com/sugarlabs/sugar-docs/issues\">report</a> any lingering links.</p>" >> index.html
+
+show-green "Done"


chimosky:
Reviewed, any particular reason why you included `python3-decorator` as one of the packages to install?

srevinsaju:
The [Archlinux PKGBUILD](https://github.com/archlinux/svntogit-community/blob/packages/sugar-toolkit-gtk3/trunk/PKGBUILD) shows that `sugar-toolkit-gtk3` has a direct dependency on `python-decorator` which translates to `python3-decorator` on ubuntu. Let me know if its unnecessary

srevinsaju:
> The instruction for using `actions/checkout@v2` says;
> "If you're using actions/checkout@v2 you must set persist-credentials to false in most cases for the deployment to work correctly.", I'm guessing this is one of the few cases where not setting `persist-credentials` doesn't affect the deployment.

Yep, until now its safe. You can compare https://github.com/srevinsaju/sugar-toolkit-gtk3/tree/master and https://github.com/srevinsaju/sugar-toolkit-gtk3/tree/gh-pages for a test run of a successful deployment

.github/workflows/docs.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: actions/checkout@v2
+        
+      - name: Prepare apt repositories
+        run: | 
+          sudo sed -i '/deb-src/s/^# //' /etc/apt/sources.list
+          sudo apt update
+
+      - name: Install sugar-toolkit-gtk3
+        run: |
+          # enable deb-src repositories
```
codewiz, line 21: This comment seems to belong with this line:
  sudo sed -i '/deb-src/s/^# //' /etc/apt/sources.list

.github/workflows/docs.yml
```diff
@@ -0,0 +1,47 @@
...
+          sudo apt install python-is-python3 python3-cairo python3-gi python3-gi-cairo python3-dateutil python3-decorator \
+            gir1.2-telepathyglib-0.12 gir1.2-webkit2-4.0 python3-empy gir1.2-sugarext-1.0
+          ./autogen.sh --with-python2
+          make
+          ./autogen.sh --with-python3
+          make
+          
+      - name: Install Sphinx
+        run: |
+          sudo apt-get -qqy install sphinx-common
```
codewiz, line 33: apt-get -> apt

srevinsaju:
@tchx84 could you check out this PR?

For context: https://developer.sugarlabs.org/sugar3 was deployed on freedom, and was not getting rebuilt. Similar to [www-sugarlabs](https://github.com/sugarlabs/www-sugarlabs).

Some more information:
* http://lists.sugarlabs.org/archive/systems/2021-January/006216.html
* http://lists.sugarlabs.org/archive/systems/2020-October/006195.html

If this works out well, we would:
* Fix this: http://lists.sugarlabs.org/archive/systems/2020-November/006210.html

.github/workflows/docs.yml
```diff
@@ -0,0 +1,47 @@
...
+          sudo sed -i '/deb-src/s/^# //' /etc/apt/sources.list
+          sudo apt update
+
+      - name: Install sugar-toolkit-gtk3
+        run: |
+          # enable deb-src repositories
+          
+          sudo apt build-dep sugar-toolkit-gtk3
+          sudo apt install python-is-python3 python3-cairo python3-gi python3-gi-cairo python3-dateutil python3-decorator \
+            gir1.2-telepathyglib-0.12 gir1.2-webkit2-4.0 python3-empy gir1.2-sugarext-1.0
```
codewiz, line 25: Shouldn't `apt build-dep sugar-toolkit-gtk3` install everything we need to build from sources? Then why do we need to install these additional packages manually?

I would use either one method or the other, but not both. Since the build deps of sugar-toolkit-gtk3 will likely drift from what was included with the build machine (ubuntu 20.04), listing the deps explicitly seems better.

codewiz (changes requested):

.github/workflows/docs.yml
srevinsaju: `apt` does not have a stable CLI interface. See [this askubuntu post](https://askubuntu.com/questions/990823/apt-gives-unstable-cli-interface-warning)

.github/workflows/docs.yml
srevinsaju: yea, will remove it.

.github/workflows/docs.yml
srevinsaju: We do not have a list of dependencies used by sugar. There were attempts to create one, but it never worked. See https://github.com/sugarlabs/sugar/issues/884#issuecomment-561384226 for some @quozl's comment during my last GCI. (reading that thread makes me regret for being dumb that time). 

We need to install the additional packages, because `build-dep` only fetches the build time dependency. To run `sugar-toolkit-gtk3`, we need to get some `runtime` dependencies. These runtime dependencies will be used to generate links to other API documentations like [python-gi documentation](https://lazka.github.io/pgi-docs) iirc. But, that becomes some extra work to keep the build time deps updated I guess :smile:. I am aware of all build time and runtime dependencies used on [Archlinux](https://aur.archlinux.org/packages/sugar-toolkit-gtk3-git/), might be helpful.

.github/workflows/docs.yml
srevinsaju: Fixed it in https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/455/commits/81e29eee2fbf885c5464fec9238f609e64b1d388, the comment was redundant

chimosky:
> Yep, until now its safe. You can compare https://github.com/srevinsaju/sugar-toolkit-gtk3/tree/master and https://github.com/srevinsaju/sugar-toolkit-gtk3/tree/gh-pages for a test run of a successful deployment

Yeah I have, looks good.

.github/workflows/docs.yml
quozl: Should use `apt-get` as `apt` is an end-user tool per `apt(8)`.  But change is needed, because your use is inconsistent; you are using `apt` for build and runtime dependencies instead of `apt-get`.

.github/workflows/docs.yml
quozl: `build-dep` doesn't do the runtime dependencies.  Also we tend to be ahead of the dependencies here, so relying the current Ubuntu package (which will likely be the latest Debian Testing package) isn't the best way.

Placing a list of build and runtime dependencies in the repository could be more discoverable; I don't know what other projects use, but perhaps `debian-build-dependencies` and `debian-runtime-dependencies` might be okay.

quozl:
Overall, I like what has been done so far, but I'm worried about having references to specific downstream release versions, because of the maintenance burden this implies.

.github/workflows/docs.yml
srevinsaju: > `build-dep` doesn't do the runtime dependencies. Also we tend to be ahead of the dependencies here, so relying the current Ubuntu package (which will likely be the latest Debian Testing package) isn't the best way.
> 
> Placing a list of build and runtime dependencies in the repository could be more discoverable; I don't know what other projects use, but perhaps `debian-build-dependencies` and `debian-runtime-dependencies` might be okay.

Agreed, placing that in the repository would solve the problem. Alternatively, using Arch Linux to build the packages would also solve the problem, as we always build on the latest stuff, and deps are always upto date. We can do either of them.

quozl: Up to you what to use.  I'm not expecting to maintain this; you've demonstrated the skills and capacity, so I'll leave it in your hands.  That said, if we do have dependency lists in the source repository, we might also change our documentation to refer to them instead of using [the `build-dep` hack](https://github.com/sugarlabs/sugar/blob/master/docs/development-environment.md#native-sugar).

srevinsaju:
Updated the branch to use `archlinux:latest` docker container. Added a shell script `ci/docs.sh` to automatically generate documentation to a `deploy` directory. The benefits are
* The container used is `archlinux:latest`. 
* We can reproduce the builds on any `archlinux:latest` docker container by running `ci/docs.sh`. This would help debug, as well as deploy it to any Continuous Integration platform, in any case such as: GitHub Actions becomes EOL in future, etc.
* Higher maintainability, we needn't rewrite the CI in future, if we decide to self host, etc.
* Uses `sugar-toolkit-gtk3-git` from the AUR, which is the latest possible package on any linux distro.
* Dependencies are directly derived from the corresponding package, aka, `PKGBUILD`

.github/workflows/docs.yml
```diff
@@ -0,0 +1,40 @@
...
+name: Documentation
+on:
+  push:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    container:
+      image: 'archlinux:latest'
```
codewiz, line 11: I'm confused: why do we have `ubuntu-20.04` above and `archlinux:latest` here?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,40 @@
...
+
+    steps:
+      - name: Install dependencies
+        run: |
+          # assume yes and do not ask for confirmation when installing deps using pacman
+          yes | pacman -Syu base-devel sudo git wget curl python-sphinx rsync --noconfirm
```
codewiz, line 17: `--noconfirm` alone should be enough, why do you also need the pipe with yes?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+
+      - name: Create builduser
+        run: |
+          # A user is required so that the docs are not built as root
+          useradd builduser -m # Create the builduser
+          passwd -d builduser # Delete the buildusers password
```
codewiz, line 27: why is this needed? sudo from root to a user won't ask for a password.

ci/docs.sh
```diff
@@ -0,0 +1,49 @@
...
+if command -v makepkg &> /dev/null
+then
+    show-green "Cloning sugar-toolkit-gtk3-git from AUR"
+    mkdir .cache
+    cd .cache 
+    git clone https://aur.archlinux.org/sugar-toolkit-gtk3-git.git
+    cd sugar-toolkit-gtk3-git
+    makepkg -sif --noconfirm 
+    show-green "Installing optional dependencies"
+    yes | sudo pacman -S --needed $(cat .SRCINFO| grep 'optdepends' | grep -o '= .*:' | sed 's,= ,,g' | sed 's,:,,g') --noconfirm
```
codewiz, line 25: same here: yes is probably unnecessary.

codewiz (changes requested):

.github/workflows/docs.yml
srevinsaju: ubuntu-20.04 is the host on which `archlinux:latest`, a docker container runs. There are options on which host can be chosen:
* ubuntu-16.04
* ubuntu-18.04
* ubuntu-20.04
* windows-latest
* macos-latest
Ubuntu is the only linux based VM available. And then we use docker to pull `archlinux:latest`

ci/docs.sh
srevinsaju: See https://unix.stackexchange.com/a/293537

.github/workflows/docs.yml
srevinsaju: We need `sudo` later in the script, in `makepkg` where it will install some deps from the `PKGBUILD`

.github/workflows/docs.yml
srevinsaju: https://unix.stackexchange.com/a/293537

.github/workflows/docs.yml
srevinsaju: To simplify a bit more, those lines expand to:
```bash

# set up the VM
az create path/to/vm/config   # set up a VM on Azure (because, its Microsoft stuff, implicit, done by GitHub interally)

ssh runner@$VM_IP_ADDR  # internal VM IP Address >>

sudo apt-get install docker  # iirc

docker run -it archlinux:latest 
```

srevinsaju:
Looking forward for a merge @quozl :smile:

quozl:
You asked @tchx84 earlier for review, you happy to proceed without?

srevinsaju:
Yea, that was because no one was reviewing this PR that time, 😂

srevinsaju:
One more final step. `Settings ->  GitHub Pages -> custom_domain = developer.sugarlabs.org  && branch = gh-pages`

srevinsaju:
Looks like it went well https://github.com/sugarlabs/sugar-toolkit-gtk3/tree/gh-pages/sugar3

srevinsaju:
Let me also know once you are done with that, so that I can update the DNS to point to `sugarlabs.github.io.`

quozl:
Changed.  Proceed.  Sorry, was breakfasting.

srevinsaju:
Ok. Thanks! Pushing now.

srevinsaju:
Done. http://developer.sugarlabs.org/sugar3/

quozl:
GitHub UI continues to say "Enforce HTTPS — Unavailable for your site because your domain is not properly configured to support HTTPS ", yet https://developer.sugarlabs.org/sugar3/ does work.

srevinsaju:
> GitHub UI continues to say "Enforce HTTPS — Unavailable for your site because your domain is not properly configured to support HTTPS ", yet https://developer.sugarlabs.org/sugar3/ does work.

Probably it might be available after LetsEncrypt TLS Certificate is provisioned. It might take a few minutes.

quozl:
Yes, it has changed; "Enforce HTTPS — Not yet available for your site because the certificate has not finished being issued.
Please allow 24 hours for this process to complete."

PR #459, opened 2021-11-21
ci: change path of placing the index.html for developer.sugarlabs.org

srevinsaju:
The index.html page was supposed to be placed in the deploy folder which would
then be staged and deployed to github actions.


srevinsaju:
I see that the CI is failing, it looks like it needs an update. I will fix it shortly

PR #460, opened 2021-11-21
ci: add sphinxcontrib.napoleon for parsing sugar docs

srevinsaju:
sphinxcontrib.napolean is not provided any more by the default
sphink package, so we will need to manually add them


srevinsaju:
Ready to merge, this will fix the CI.

PR #470, opened 2024-01-18
Fix issue with Buddy.props.key being bytes

chimosky:
Fixes #443 

Buddy.props.key is supposed to be a str, but is stored as a dbus.ByteArray before being sent to dbus, decoding it after the call to GetProperties fixes the issue of it being bytes.

Activities like maze that use Buddy.props.key needed to encode the received key before comparison worked.

@quozl kindly review.

Tested without `key = str(key.encode())` in maze activity and collaboration works as expected.


quozl:
Looks fine.  Guess you'll need to version the fix in Maze now.

chimosky:
Done.

PR #473, opened 2024-04-04
Add weblate sync workflow

chimosky:
This will automatically open a PR whenever weblate pushes changes to weblate-sync.

### DIFF
--- a/.github/workflows/weblate-sync.yaml
+++ b/.github/workflows/weblate-sync.yaml
@@ -1,0 +1,14 @@
+name: Weblate - Create PR when weblate pushes changes
+on:
+  push:
+    branches:
+      - weblate-sync
+jobs:
+  weblateSync:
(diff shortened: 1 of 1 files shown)


quozl:
File name has syc instead of sync?

Is anyone looking after the CI/CD checks, or should we really disable them because they don't have a champion?

chimosky:
> File name has syc instead of sync?
> 
> Is anyone looking after the CI/CD checks, or should we really disable them because they don't have a champion?

Fixed the filename, I'll look at the error with building the doc.

chimosky:
@quozl we should merge this as it has nothing to do with the documentation workflow that's failing, I looked into it before and just started looking at it again.

quozl:
I'm okay for you to merge this.  I would do it myself but I'm not confident I know enough to support it.

PR #474, opened 2024-05-28
Activity.py Porting from GTK3 to GTK4

inkerton:
Ported the file according to the porting commands.
Updates some imports in activity.py


quozl:
No useful commit messages.  Looked at files changed, but didn't know what was the goal or if it had been reached.  I suggest breaking the work into intentional pieces.

PR #475, opened 2024-10-06
drop Python 2.3 fallback

a-detiste:


walterbender:
This probably makes sense, but there is no associated ticket or explanation for this change.

a-detiste:
Hi, I'm just testing the waters to see if Python2 support can be wholefully removed.

Asking for a ticket means I know I won't try to get further.

https://wiki.debian.org/Python3-six-removal

a-detiste:
Sorry I think I'm being abrupt here assuming unsaid thing :-|

chimosky:
> Sorry I think I'm being abrupt here assuming unsaid thing :-|

It's totally okay, I do agree with the change.

chimosky:
Merged by hand, thank you!

PR #485, opened 2025-03-16
Add GitHub issue template for DMP 2025

pikurasa:
This is the GitHub issue template for this year's DMP projects.

I plan on adding https://github.com/sugarlabs/GSoC/blob/master/Ideas-2025.md#gtk4-exploration via the template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/DMP_2024.yml
+++ b/.github/ISSUE_TEMPLATE/DMP_2024.yml
@@ -1,0 +1,234 @@
+name: DMP 2025 Project Template
+description: List a new project for Dedicated Mentoring Program (DMP) 2025 
+title: "[DMP 2025]: "
+labels: ["DMP 2025"]
+body:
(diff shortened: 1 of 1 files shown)


chimosky:
If we're not working on the toolkit in DMP then this should be in the repo that we'll be working on, IIRC.

pikurasa:
> If we're not working on the toolkit in DMP then this should be in the repo that we'll be working on, IIRC.

For this particular DMP issue, would that be initiating a new repo and putting it there? If so, what do we name it?

And if there's already a repo, please point me there.

chimosky:
> > If we're not working on the toolkit in DMP then this should be in the repo that we'll be working on, IIRC.
> 
> For this particular DMP issue, would that be initiating a new repo and putting it there? If so, what do we name it?
> 
> And if there's already a repo, please point me there.

This repo works fine, we can use separate branches when we need to.

PR #487, opened 2025-04-02
Detects Python 2 Activities_  fixes /sugar #991

ahmadtariq1:
**Changes in This PR**
In sugar-toolkit-gtk3’s activityfactory.py, I’ve implemented:

- Handling FileNotFoundError: Added a try-except block around the subprocess.Popen call to catch FileNotFoundError.
- Detection of "sugar-activity": When the exception occurs and involves "sugar-activity,", error is logged to the activity log.
- Communication of launch failure.

**Remaining Work**
The following requirements are addressed in a separate PR in Sugar:

Reducing the launcher animation timeout to ~10 seconds.
Displaying error message in launcher.

**Testing**
Tested on Fedora 41 with Sugar 0.121.
Used calculate-activity(version 44) to verify:
The FileNotFoundError is caught and logged.

**Notes**
Please review the companion PR in the Sugar repo for the complete view of the fix.

@chimosky kindly review this.

### DIFF
--- a/src/sugar3/activity/activityfactory.py
+++ b/src/sugar3/activity/activityfactory.py
@@ -37,6 +37,8 @@ from errno import EEXIST, ENOSPC
 import os
 import subprocess
 
+
+
 _SHELL_SERVICE = 'org.laptop.Shell'
 _SHELL_PATH = '/org/laptop/Shell'
 _SHELL_IFACE = 'org.laptop.Shell'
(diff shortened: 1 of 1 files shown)


src/sugar3/activity/activityfactory.py
```diff
@@ -217,25 +219,59 @@ def _create_activity(self):
             self._service_name, self._handle.activity_id,
             reply_handler=self._no_reply_handler,
             error_handler=self._notify_launch_error_handler)
-
+
```
chimosky, line 222: An unnecessary change.

src/sugar3/activity/activityfactory.py
```diff
@@ -345,4 +381,4 @@ def error_handler_cb(error):
         # main window, see http://bugs.sugarlabs.org/ticket/1447#comment:19
         shell.NotifyLaunchFailure(activity_id,
                                   reply_handler=reply_handler_cb,
-                                  error_handler=error_handler_cb)
+                                  error_handler=error_handler_cb)
```
chimosky, line 384: Please remove the No new line at EOF, you can do so with `echo >> activityfactory.py`.

chimosky (commented):
Reviewed, not tested.

src/sugar3/activity/activityfactory.py
quozl: Also see if you can figure out how it happened; we don't want unnecessary changes, it consumes reviewing time, and makes `git bisect` and `git blame` slightly more interesting than we can emotionally deal with.

quozl:
* [ ] commit message is almost empty, please rewrite it and force push; include problem being solved (not links), and how you solve it.  See [Making Commits](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits).
* [ ] using D-Bus to send back the launch failure seems a bit odd, please explain why you chose that,
* [ ] the test for sugar-activity in the error is frankly strange, and will also trigger if sugar-activity3 is in the error,
* [ ] why not check for `sugar-activity ` in `exec`?

PR #497, opened 2026-01-17
fix: used get_target() to ensure correctness of drag and drop option

### LINKED ISSUE #496: Drag and drop fails due to missing get_target()
Paint activity triggers an error while drag and drop, originating from sugar3/graphics/colorbutton.py. This caused by an outdated selection target api.

this issue was originally reported in sugarlabs/paint-activity#43

Akshay-krish:
This PR fixes the AttributeError: 'SelectionData' object has no attribute 'target' reported in sugarlabs/paint-activity#43

The error occurs because selection_data.target is not available in GTK3. This change replaces it with selection_data.get_target(), ensuring the drag-and-drop of colors works as expected.

fixes #496

### DIFF
--- a/src/sugar3/graphics/colorbutton.py
+++ b/src/sugar3/graphics/colorbutton.py
@@ -243,7 +243,8 @@ class _ColorButton(Gtk.Button):
     def __drag_data_get_cb(self, widget, context, selection_data, info, time):
         data = struct.pack('=HHHH', self._color.red, self._color.green,
                            self._color.blue, 65535)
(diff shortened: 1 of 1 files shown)


quozl:
What about the FIXME comment on line 73, are you now claiming that drag and drop works, if so what is your test procedure?

quozl:
Updated issue with reproducer.  Tested on Ubuntu 24.04.

It solves one traceback but causes another.  It seems you haven't actually worked on the problem.  Please test your changes before making a pull request, to ensure they are complete and do not cause further problems.

PR #502, opened 2026-02-12
chore: Fix CI, documentation build, and modernization

Inuth0603:
This PR resolves CI build failures caused by the runner upgrade to Ubuntu 24.04 (Noble) and fixes various compilation errors.

### Changes
* **CI/Workflow:** Removed deprecated `python-all-dev` (replaced with `python3-all-dev`), added `autotools` dependencies, and removed unnecessary `builduser` permission workarounds in `docs.yml`.
* **Build Scripts:** Updated `make-doc.sh` to use absolute paths to prevent directory traversal issues and removed `sudo` from `ci/docs.sh` to fix permission errors.
* **C Compilation:** Fixed `const char*` warnings in `sugar-fatattr.c` and deprecated GDK calls in `sugar-key-grabber.c` (required for modern GCC/GTK versions).
* **Config:** Relaxed `configure.ac` for Autoconf 2.69 compatibility and updated `doc/conf.py` to support Sphinx 7.x.

### Verification
Local build passes. The CI environment should now be able to install dependencies and compile the C extensions on Ubuntu 24.04 without "Package not found" errors.

### DIFF
--- a/src/sugar3/sugar-fatattr.c
+++ b/src/sugar3/sugar-fatattr.c
@@ -33,7 +33,7 @@
 #include <string.h>
 
 // This function is wrapped by getattrs/setattrs
-static int _ioctl_attrs(char *file, __u32 *attrs, int ioctlnum, char *verb)
+static int _ioctl_attrs(const char *file, __u32 *attrs, int ioctlnum, char *verb)
 {
     int fd;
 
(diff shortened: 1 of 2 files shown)


src/sugar3/sugar-fatattr.c
chimosky: Did building the toolkit throw any errors here?

chimosky (commented):
Would be nice to split the commits and keep the CI changes to one commit, and keep any non CI related changes to another.

src/sugar3/sugar-fatattr.c
Inuth0603: Yes, the build would fail or produce significant warnings because set_hidden_attrib was assigning a const char* to a non-const char*, discarding the qualifier. With GNOME_COMPILE_WARNINGS(maximum) enabled in configure.ac, this is strictly flagged. I've separated the CI/build infrastructure changes from the code fixes as requested.

